PR #5, opened 2020-10-06
Added host in redirect url

adeelehsan:
Backends returns redirect url as path(`/dasboard`) instead of absolute url. Appended lms base url for redirection


waheedahmed:
Can we close this out if not needed anymore?

PR #6, opened 2020-10-07
Adds Reset password functionality.

adeelkhan:
Adds pages for;
- Password reset form.
- Successful reset message
- Invalid token message.

Adds sagas/service code for;
- Token validation
- Password reset

VAN-77

### DIFF
--- a/src/reset-password/InvalidToken.jsx
+++ b/src/reset-password/InvalidToken.jsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { Hyperlink } from '@edx/paragon';
+
+import Alert from '../logistration/Alert';
+
+const InvalidTokenMessage = () => {
+  const loginPasswordLink = (
+    <Hyperlink destination="/login">
+      <FormattedMessage
+        id="logistration.forgot.password.confirmation.support.link"
+        defaultMessage="sign-in"
+        description="link text used in message: logistration.reset.password.request.invalid.token.description.message link 'sign-in.'"
+      />
+    </Hyperlink>
+  );
+
+  const forgotPassword = (<strong>Forgot Password</strong>);
+  return (
+    <div className="d-flex justify-content-center reset-password-container">
+      <div className="d-flex flex-column" style={{ width: '400px' }}>
+        <div className="form-group">
+          <div className="text-center mt-3">
+            <Alert className="alert-danger mt-n2">
+              <h4 style={{ color: '#a0050e' }}>
+                <FormattedMessage
+                  id="logistration.reset.password.request.invalid.token.header.message"
+                  defaultMessage="Invalid Password Reset Link"
+                  description=""
+                />
+              </h4>
+              <FormattedMessage
+                id="logistration.reset.password.request.invalid.token.description.message"
+                defaultMessage="This password reset link is invalid. It may have been used already. To reset your password, go to the {loginPasswordLink} page and select {forgotPassword}"
+                description=""
+                values={{
+                  forgotPassword,
+                  loginPasswordLink,
+                }}
+              />
+            </Alert>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default InvalidTokenMessage;
--- a/src/reset-password/ResetPasswordPage.jsx
+++ b/src/reset-password/ResetPasswordPage.jsx
@@ -1,0 +1,164 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { Button, Input, ValidationFormGroup } from '@edx/paragon';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import messages from './messages';
+import { resetPassword, validateToken } from './data/actions';
+import { resetPasswordResultSelector } from './data/selectors';
+import InvalidTokenMessage from './InvalidToken';
+import ResetSuccessMessage from './ResetSuccess';
+import ResetFailureMessage from './ResetFailure';
+import Spinner from './Spinner';
+
+
+const ResetPasswordPage = (props) => {
+  const { intl } = props;
+
+  const [newPasswordInput, setNewPasswordValue] = useState('');
+  const [confirmPasswordInput, setConfirmPasswordValue] = useState('');
+  const [passwordValid, setPasswordValidValue] = useState(true);
+  const [passwordMatch, setPasswordMatchValue] = useState(true);
+
+  const handleNewPasswordChange = (e) => {
+    const newPassword = e.target.value;
+    setNewPasswordValue(newPassword);
+    const isValid = newPassword.length >= 8 && newPassword.match(/\d+/g);
+    setPasswordValidValue(isValid !== false);
+  };
+  const handleConfirmPasswordChange = (e) => {
+    const confirmPassword = e.target.value;
+    setConfirmPasswordValue(confirmPassword);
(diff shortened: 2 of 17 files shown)


src/reset-password/data/sagas.js
```diff
@@ -0,0 +1,51 @@
...
+} from './actions';
+
+import { validateToken, resetPassword } from './service';
+
+
+// Services
+export function* handleValidateToken(action) {
+  try {
+    yield put(validateTokenBegin());
+    const { isValid } = yield call(validateToken, action.payload.token);
```
waheedahmed, line 22: The returned key is `is_valid` or we should change the endpoint to return `isValid` otherwise it will always be undefined.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,178 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+
+import messages from './messages';
+import { resetPassword, validateToken } from './data/actions';
+import { resetPasswordResultSelector } from './data/selectors';
+import InvalidTokenMessage from './InvalidToken';
+import ResetSuccessMessage from './ResetSuccess';
+
+const Spinner = () => (
```
waheedahmed, line 16: Create it's own component so that we can use this in other components as well.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,178 @@
...
+  <div className="d-flex justify-content-center reset-password-container">
+    <div className="d-flex flex-column" style={{ width: '400px' }}>
+      <div className="form-group">
+        <div className="d-flex flex-column align-items-start">
+          Validating Token
```
waheedahmed, line 21: Remove the text and spinner should be in aligned center both vertically and horizontally.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,178 @@
...
+      </div>
+    </div>
+  </div>
+);
+
+const InvalidTokenError = () => (
+  <div className="d-flex justify-content-center reset-password-container">
+    <div className="d-flex flex-column" style={{ width: '400px' }}>
+      <div className="form-group">
+        <div className="d-flex flex-column align-items-start">
```
waheedahmed, line 33: Move these parent divs into its component `InvalidTokenMessage`, same for the `ResetSuccessMessage` component.

waheedahmed (commented):
Generally looks good, also add the test cases and let me know when it's ready for the next pass.

src/reset-password/ResetPasswordPage.jsx
waheedahmed: Here is an example of how it's implemented in accounts MFE [PageLoading.jsx](https://github.com/edx/frontend-app-account/blob/8020ff58b8e8e8b73b066a012ab2da65213d394b/src/account-settings/PageLoading.jsx#L6)

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,143 @@
...
+      return (
+        <>
+          <div className="d-flex justify-content-center reset-password-container">
+            <div className="d-flex flex-column" style={{ width: '400px' }}>
+              <form className="m-0">
+                <div className="form-group">
+                  <h3 className="text-center mt-3">
+                    {intl.formatMessage(messages['logisration.reset.password.page.heading'])}
+                  </h3>
+                  <p className="mb-4">
```
waheedahmed, line 70: Add `mt-4` and decrease the margin bottom to 2. With this the layout looks little odd.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,143 @@
...
+      };
+      props.resetPassword(formPayload, props.token);
+    }
+  };
+
+  if (props.token_status === 'pending') {
+    const { token } = props.match.params;
+    if (token) {
+      props.validateToken(token);
+      return <Spinner text='Token validation in progress .. ' />;
```
waheedahmed, line 53: Do we need to format this message?

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -0,0 +1,143 @@
...
+  const handleConfirmPasswordChange = (e) => {
+    const confirmPassword = e.target.value;
+    setConfirmPasswordValue(confirmPassword);
+    setPasswordMatchValue(confirmPassword === newPasswordInput);
+  };
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+    if (newPasswordInput === '') {
+      setPasswordValidValue(false);
```
waheedahmed, line 38: You also need to return after this, otherwise it will call the action because the `passwordValid` state will stay `true` until the the react reloads the component. Or change it to if/else.

src/reset-password/ResetPasswordPage.jsx
adeelkhan: Done

src/reset-password/ResetPasswordPage.jsx
adeelkhan: I have changed it to resemble Forgot Password page now and now look and feel is more consistent with it.

src/reset-password/Spinner.jsx
```diff
@@ -0,0 +1,18 @@
...
+        <div className="d-flex flex-column align-items-start">
+        <h3 className="text-center mt-3" >{props.text ? props.text: ''} 
+          <FontAwesomeIcon icon={faSpinner} spin />
+        </h3>   
+        </div>
+      </div>
+    </div>
+  </div>
+);
+export default Spinner;
```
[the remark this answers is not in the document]
adeelkhan: Done

src/reset-password/Spinner.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const textMessage = props.text;
+  return (
+    <div className="d-flex justify-content-center reset-password-container">
+      <div className="d-flex flex-column" style={{ width: '400px' }}>
+        <div className="form-group">
+          <div className="d-flex flex-column align-items-start">
+            <h3 className="text-center mt-3">
+              <FormattedMessage
+                id="logistration.reset.password.request.token.validation.message"
+                defaultMessage="{textMessage}"
```
waheedahmed, line 17: I believe it will not get extracted for translations in this way.

src/reset-password/InvalidToken.jsx
```diff
@@ -0,0 +1,49 @@
...
+    </Hyperlink>
+  );
+
+  const forgotPassword = (<strong>Forgot Password</strong>);
+  return (
+    <div className="d-flex justify-content-center reset-password-container">
+      <div className="d-flex flex-column" style={{ width: '400px' }}>
+        <div className="form-group">
+          <div className="text-center mt-3">
+            <Alert className="alert-warning mt-n2">
```
waheedahmed, line 24: Change the class to `alert-danger` to match the old FE message.

waheedahmed (approved):
Couple of comments, otherwise looks good 👍 

Also rebase your branch with latest master, I have changed some `codecov` configurations.

src/reset-password/Spinner.jsx
adeelkhan: Iam not sure ... this can be checked.

waheedahmed: IN this way the `textMessage` is treated as a variable and its value won't be available to the translators. Same like the inserting some values into a string using `interpolate()` and `.format()` functions.

PR #9, opened 2020-10-20
Enable forgot password functinality for secondary email.

adeelkhan:
We need to enhance forgot password feature to enable user to get reset password link on his/her secondary email address if set inside db. User should be able to get reset link on secondary email address and use it to recover account incase he/she is somehow blocked from using primary email address.
VAN-18


adeelehsan (approved):
looks good 👍

uzairr (approved):
LGTM 👍

PR #10, opened 2020-10-22
added redirection for registration

adeelehsan:
VAN-7


src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+    };
+  });
+
+  it('should match default section snapshot', () => {
+    const tree = renderer.create(reduxWrapper(<IntlRegistrationPage {...props} />));
+    expect(tree.toJSON()).toMatchSnapshot();
+  });
+
+  it('should match url after redirection', () => {
+    const dasboardUrl = 'http://test.com/dashboard/';
```
waheedahmed, line 53: Can we change this url to something else? Because the default value is also dashboard URL, it would be great if we test that case as well.

src/logistration/RegistrationPage.jsx
```diff
@@ -40,6 +43,7 @@ class RegistrationPage extends React.Component {
 
   handleSubmit = (e) => {
     e.preventDefault();
+    const params = getQueryParameters();
```
waheedahmed, line 46: It's deprecated, instead use `URL.searchParam`, check this for more details https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams. Also replace it in `LoginPage.jsx`.

src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+      config: {
+        ENVIRONMENT: 'production',
+        LANGUAGE_PREFERENCE_COOKIE_NAME: 'yum',
+      },
+      messages: {
+        'es-419': {},
+        de: {},
+        'en-us': {},
+      },
+    });
```
waheedahmed, line 40: Why we need all this here?

waheedahmed (commented):
Few minor comments, otherwise looks goos.

src/logistration/tests/RegistrationPage.test.jsx
adeelehsan: we need i18 configuration because we are using `getLocale and getCountryList ` methods. If we don't add the configuration it would throw the error getLocale called before the configure.

src/logistration/RegistrationPage.jsx
waheedahmed: I believe you forgot to change it here?

src/logistration/tests/RegistrationPage.test.jsx
waheedahmed: This comment also seems unaddressed?

waheedahmed (approved):
Looks great 👍, thank you!

PR #11, opened 2020-10-23
Add tests for login functionality

zainab-amir:
Tests added for:
- LoginPage
- LoginHelpLinks

[VAN-60](https://openedx.atlassian.net/browse/VAN-60)

### DIFF
--- a/src/logistration/tests/LoginHelpLinks.test.jsx
+++ b/src/logistration/tests/LoginHelpLinks.test.jsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import { IntlProvider } from '@edx/frontend-platform/i18n';
+import { mount } from 'enzyme';
+
+import LoginHelpLinks from '../LoginHelpLinks';
+
+describe('LoginHelpLinks', () => {
+  let props = {};
+  const otherSignInIssues = 'https://support.edx.org/hc/en-us/sections/115004153367-Solve-a-Sign-in-Problem';
+
+  const reduxWrapper = children => (
+    <IntlProvider locale="en">
+      {children}
+    </IntlProvider>
+  );
+
+  it('renders help links on button click', () => {
+    props = {
+      ...props,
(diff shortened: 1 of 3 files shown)


src/logistration/tests/LoginHelpLinks.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+describe('LoginHelpLinks', () => {
+  let props = {};
+
+  it('should match login page help links snapshot', () => {
+    props = {
+      ...props,
+      page: 'login',
+    };
+    const tree = renderer.create(
+      <IntlProvider locale="en">
```
adeelehsan, line 16: Please move that code to `beforeEach`

src/logistration/tests/LoginHelpLinks.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should match forget password page help links snapshot', () => {
+    props = {
+      ...props,
+      page: 'forget-password',
+    };
+
+    const tree = renderer.create(reduxWrapper(<LoginHelpLinks {...props} />)).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
waheedahmed, line 35: I believe these tests are unnecessary and only verifying the `Need other help signing in?` button which is already been tested in `LoginPage.test.jsx` and `ForgotPasswordPage.test.jsx`.

Instead of this I believe we should test; on expanding the it renders the right links for both pages.

waheedahmed (commented):
One comment otherwise looks good 👍

src/logistration/tests/LoginHelpLinks.test.jsx
zainab-amir: got it. Will update these tests

waheedahmed (approved):
Looks great 👍

PR #13, opened 2020-10-29
Adds error handling to Login page.

adeelkhan:
This patch handles 400 error for login
page and displays backend error to learner.

VAN-105

Screen Shot 2020-11-02 at 4 27 54 PM

### DIFF
--- a/src/logistration/LoginFailure.jsx
+++ b/src/logistration/LoginFailure.jsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
+import { Alert, Hyperlink } from '@edx/paragon';
+
+const processLink = (link) => {
+  let matches;
+  link.replace(/(.*)<a href=["']([^"']*).*>([^<]+)<\/a>(.*)/g, function () { // eslint-disable-line func-names
+    matches = Array.prototype.slice.call(arguments, 1, 5); // eslint-disable-line  prefer-rest-params
+  });
+  return matches;
+};
+
+const LoginFailureMessage = (props) => {
+  const Link = (args) => (
+    <>
+      {args.beforeLink}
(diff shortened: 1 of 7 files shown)


src/logistration/LoginFailure.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
+import { Alert, Hyperlink } from '@edx/paragon';
+
+const processLink = (link) => {
+  let matches;
+  link.replace(/(.*)<a href=["']([^"']*).*>([^<]+)<\/a>(.*)/g, function () { // eslint-disable-line func-names
```
adeelehsan, line 9: Can we move that expression to constant's or at least assign to a variable ?

src/logistration/LoginPage.jsx
```diff
@@ -13,6 +13,26 @@ import { DEFAULT_REDIRECT_URL } from './data/constants';
 import { loginRequestSelector, thirdPartyAuthContextSelector } from './data/selectors';
 import LoginHelpLinks from './LoginHelpLinks';
 import RedirectLogistration from './RedirectLogistration';
+import LoginFailureMessage from './LoginFailure';
+
+
+const LoginRedirect = (props) => {
```
adeelehsan, line 19: That code is already there. Please rebase.

adeelehsan (changes requested):
looks good only few nit's

src/logistration/LoginPage.jsx
adeelkhan: Done!

src/logistration/LoginFailure.jsx
adeelkhan: This has single usage, so i kept it like this. Name of the function do explain what the regex is doing.

adeelehsan (approved):

PR #16, opened 2020-11-04
Fetch third party auth context

zainab-amir:
Fetch current provider details, social auth and secordary auth provider
details from edx-platform api.

VAN-92


src/logistration/LoginPage.jsx
```diff
@@ -48,14 +49,22 @@ class LoginPage extends React.Component {
...
+    const payload = {
+      redirect_to: params.get('next') || DEFAULT_REDIRECT_URL,
+    };
+    this.props.getThirdPartyAuthContext(payload);
+  }
+
   handleSubmit = (e) => {
     e.preventDefault();
-    const params = getQueryParameters();
+    const params = (new URL(document.location)).searchParams;
```
adeelehsan, line 62: just for clarification why we are removing `getQueryParameters`

src/logistration/data/service.js
```diff
@@ -42,3 +42,25 @@ export async function login(creds) {
...
+      `${getConfig().LMS_BASE_URL}/api/third_party_auth_context`,
+      { params: urlParams },
+      requestConfig,
+    )
+    .catch((e) => {
+      throw (e);
+    });
+
+  return {
+    thirdPartyAuthContext: data,
```
adeelehsan, line 64: You have to write a response transformer as well which would convert the response to camel case.

src/logistration/tests/LoginPage.test.jsx
```diff
@@ -84,7 +84,7 @@ describe('LoginPage', () => {
       },
     });
     delete window.location;
-    window.location = { href: '' };
+    window.location = { href: '', search: '' };
```
adeelehsan, line 87: Why we are adding search in windows.location ?

src/logistration/LoginPage.jsx
zainab-amir: This is deprecated now in favour of `getQueryParameters()`

src/logistration/tests/LoginPage.test.jsx
zainab-amir: the test start failing because we added `getQueryParameters ` in `componentDidMount()`.
`getQueryParameters()` extracts query params from `global.location.search` and return those after parsing. Because this test is deleting the window.location, it was raising an error. 

I will see if we still need to set this field or not since we have moved from `getQueryParameters` to `searchParams`

src/logistration/data/service.js
```diff
@@ -42,3 +42,25 @@ export async function login(creds) {
     success: data.success || false,
   };
 }
+
+
+export async function getThirdPartyAuthContext(urlParams) {
+  const requestConfig = {
+    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+    isPublic: true,
```
waheedahmed, line 50: No need to define this with `getHttpClient()`.

waheedahmed (approved):
Looks great, 👍 once coverage is fixed.

PR #17, opened 2020-11-04
Add backend validation on password reset page.

waheedahmed:
On old password reset page we use backend endpoint to validate password, added same functionality for the MFE.

VAN-87

### DIFF
--- a/src/reset-password/ResetPasswordPage.jsx
+++ b/src/reset-password/ResetPasswordPage.jsx
@@ -7,6 +7,7 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import messages from './messages';
 import { resetPassword, validateToken } from './data/actions';
 import { resetPasswordResultSelector } from './data/selectors';
+import { validatePassword } from './data/service';
 import InvalidTokenMessage from './InvalidToken';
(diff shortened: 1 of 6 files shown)


src/reset-password/ResetPasswordPage.jsx
```diff
@@ -20,12 +23,37 @@ const ResetPasswordPage = (props) => {
...
   const [passwordValid, setPasswordValidValue] = useState(true);
   const [passwordMatch, setPasswordMatchValue] = useState(true);
+  const [validationMessage, setvalidationMessage] = useState('');
+
+  const validatePassword = async (password) => {
+    const requestConfig = {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+      isPublic: true,
+    };
+    const { data } = await getAuthenticatedHttpClient()
```
zainab-amir, line 33: any reason why we have not written it inside the service.js file?

waheedahmed: @zainab-amir No specific reason, I believe we can move this into service.

waheedahmed: @zainab-amir Done, kindly take another look.

zainab-amir (approved):

PR #22, opened 2020-11-06
Added saml integration for login and registration

adeelehsan:
image
image

image
image


src/logistration/RegistrationPage.jsx
```diff
@@ -142,6 +157,16 @@ class RegistrationPage extends React.Component {
   }
 
   render() {
+    if (this.state.institutionLogin) {
+      return (
+        <InstitutionLogistration
+          onSubmitHandler={this.handleInstitutionLogin}
+          secondaryProviders={this.props.thirdPartyAuthContext.secondaryProviders}
```
zainab-amir, line 164: Will this not cause issue if component has not mounted yet and you try to access it? since `thirdPartyAuthContext` will be undefined by then?

adeelehsan: Nope this code would be executed when `this.state.institutionLogin` is true. it's value by default is false and only true when we click on the button `use my uni info` and that button only appears when thirdPartyContext is not empty.

src/logistration/InstitutionLogistration.jsx
```diff
@@ -0,0 +1,89 @@
...
+  secondaryProviders: [],
+  buttonTitle: '',
+};
+const LogistrationProps = {
+  onSubmitHandler: PropTypes.func.isRequired,
+  secondaryProviders: PropTypes.arrayOf(PropTypes.shape({
+    name: PropTypes.string.isRequried,
+    loginUrl: PropTypes.string.isRequired,
+  })),
+  ButtonTitle: PropTypes.string,
```
waheedahmed, line 15: I believe it should be in camelCase `buttonTitle`?

src/logistration/LoginPage.jsx
```diff
@@ -104,6 +109,16 @@ class LoginPage extends React.Component {
   }
 
   render() {
+    if (this.state.institutionLogin) {
+      return (
+        <InstitutionLogistration
+          onSubmitHandler={this.handleInstitutionLogin}
+          secondaryProviders={this.props.thirdPartyAuthContext.secondaryProviders}
+          headingTitle="Sign in with Institution/Campus Credentials"
+          buttonTitle="Back"
```
waheedahmed, line 118: Mark these strings for translations.

src/logistration/LoginPage.jsx
```diff
@@ -119,10 +134,17 @@ class LoginPage extends React.Component {
                 First time here?<a className="ml-1" href="/register">Create an Account.</a>
               </p>
             </div>
+            <h3 className="text-left mt-3">Sign In</h3>
+            <RenderInstitutionButton
+              onSubmitHandler={this.handleInstitutionLogin}
+              secondaryProviders={this.props.thirdPartyAuthContext.secondaryProviders}
+              buttonTitle="Use my university info"
```
waheedahmed, line 141: Same here, mark this string for translations.

src/logistration/LoginPage.jsx
```diff
@@ -119,10 +134,17 @@ class LoginPage extends React.Component {
...
+            />
+            <div className="section-heading-line mb-4">
+              <h4>or sign in with</h4>
```
waheedahmed, line 144: And here as well.

waheedahmed (commented):
Most of the strings are not marked for the translations, I have added comment on few but there are more. Please mark those using `defineMessages` or `FormattedMessage`.

Also add some screenshots if possible.

src/logistration/InstitutionLogistration.jsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export const RenderInstitutionButton = props => {
```
zainab-amir, line 18: should we move the exports at the end of file?

src/logistration/tests/LoginPage.test.jsx
```diff
@@ -119,6 +122,7 @@ describe('LoginPage', () => {
       ...store,
       logistration: {
         ...store.logistration,
+        thirdPartyAuthContext: { secondaryProviders: [] },
```
zainab-amir, line 125: shouldn't we just add it inside the initialState?

src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -72,6 +74,62 @@ describe('./RegistrationPage.js', () => {
...
+
+  it('should display institution register page', () => {
+    store = mockStore({
+      ...store,
+      logistration: {
+        ...store.logistration,
+        registrationResult: {
+          success: true,
+          redirectUrl: '',
+        },
```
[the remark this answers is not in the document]
zainab-amir: @adeelehsan ^

src/logistration/tests/LoginPage.test.jsx
```diff
@@ -219,4 +219,62 @@ describe('LoginPage', () => {
...
+    store = mockStore({
+      ...store,
+      logistration: {
+        ...store.logistration,
+        loginResult: {
+          success: true,
+          redirectUrl: '',
+        },
+        thirdPartyAuthContext: {
+          providers: [],
```
zainab-amir, line 233: no need to explicitly write this. We can use the default defined above,
```javascript
thirdPartyAuthContext: {
  ...store.logistration.thirdPartyAuthContext,
  secondaryProviders: [
  {
     id: 'saml-test',
     name: 'Test University',
     loginUrl: '/dummy-auth',
     registerUrl: '/dummy_auth'
   },
}

File: src/logistration/tests/LoginPage.test.jsx [hunk 10]
```diff
@@ -219,4 +219,62 @@ describe('LoginPage', () => {
...
+          secondaryProviders: [
+            {
+              id: 'saml-test',
+              name: 'Test University',
+              loginUrl: '/dummy-auth',
+              registerUrl: '/dummy_auth',
+            },
```
zainab-amir, line 240: Can we define this as a const outside of test? This way we won't have to write repeated code for all tests that require the secondaryProviders

src/logistration/RegistrationPage.jsx
```diff
@@ -158,7 +189,12 @@ class RegistrationPage extends React.Component {
             <span className="d-block mx-auto mb-4 section-heading-line">Create an account using</span>
             <button type="button" className="btn-social facebook"><FontAwesomeIcon className="mr-2" icon={faFacebookF} />Facebook</button>
             <button type="button" className="btn-social google"><FontAwesomeIcon className="mr-2" icon={faGoogle} />Google</button>
-            <button type="button" className="btn-social microsoft"><FontAwesomeIcon className="mr-2" icon={faMicrosoft} />Microsoft</button>
+            <button type="button" className="btn-social microsoft mb-3"><FontAwesomeIcon className="mr-2" icon={faMicrosoft} />Microsoft</button>
+            <RenderInstitutionButton
+              onSubmitHandler={this.handleInstitutionLogin}
+              secondaryProviders={this.props.thirdPartyAuthContext.secondaryProviders}
+              buttonTitle={this.props.intl.formatMessage(messages['logistration.register.institution.login.button'])}
```
zainab-amir, line 196: can probably define `intl` outside return for cleaner code
```javascript
const { intl } = this.props;
```

zainab-amir (approved):
Once the code changes are addressed

src/logistration/InstitutionLogistration.jsx
```diff
@@ -0,0 +1,100 @@
...
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import PropTypes from 'prop-types';
+import { Button } from '@edx/paragon';
+import messages from './messages';
+
+const LogistrationDefaultProps = {
+  secondaryProviders: [],
+  buttonTitle: '',
+};
+const LogistrationProps = {
```
waheedahmed, line 11: We should move this to the bottom where we are actually defining the `propTypes` for the components.

waheedahmed (approved):
👍 once all translations related comments are addressed.

PR #23, opened 2020-11-09
modify reset password message and background color

uzairr:
ACs:
1. the message copy for MFE needs to be updated to match the hyperlink text same as old FE.
2. Message placeholder should be green for success.

VAN-106


src/reset-password/ResetSuccess.jsx
```diff
@@ -8,7 +8,7 @@ const ResetSuccessMessage = () => {
     <Hyperlink destination="/login">
```
waheedahmed, line 8: Use `Alert.Link`.

waheedahmed (commented):
One suggestion, otherwise looks good.

waheedahmed (approved):

PR #26, opened 2020-11-12
add social auth

zainab-amir:
Add third party auth providers

- Added TPA providers to login page
- Added error component to deal with third party auth error

Following are generated using the prod and stage TPA context.

Screenshot 2020-11-12 at 5 46 45 PM
Screenshot 2020-11-12 at 5 49 28 PM
Screenshot 2020-11-12 at 11 04 34 AM

[VAN-92](https://openedx.atlassian.net/browse/VAN-92)

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -1,0 +1,8 @@
+// URL Paths
+export const LOGIN_PAGE = '/login';
+export const REGISTER_PAGE = '/register';
+export const RESET_PAGE = '/reset';
+export const DEFAULT_REDIRECT_URL = '/dashboard';
+
+// Constants
+export const SUPPORTED_ICON_CLASSES = ['apple', 'facebook', 'google', 'microsoft'];
--- a/src/logistration/LoginPage.jsx
+++ b/src/logistration/LoginPage.jsx
@@ -1,19 +1,19 @@
 import React from 'react';
 
 import { Button, Input, ValidationFormGroup } from '@edx/paragon';
-import { faFacebookF, faGoogle, faMicrosoft } from '@fortawesome/free-brands-svg-icons';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import { forgotPasswordResultSelector } from '../forgot-password';
 import ConfirmationAlert from './ConfirmationAlert';
 import { getThirdPartyAuthContext, loginRequest } from './data/actions';
-import { DEFAULT_REDIRECT_URL } from './data/constants';
+import { DEFAULT_REDIRECT_URL, REGISTER_PAGE } from '../data/constants';
 import { loginRequestSelector, thirdPartyAuthContextSelector } from './data/selectors';
 import LoginHelpLinks from './LoginHelpLinks';
-import RedirectLogistration from './RedirectLogistration';
 import LoginFailureMessage from './LoginFailure';
+import RedirectLogistration from './RedirectLogistration';
+import SocialAuthProviders from './SocialAuthProviders';
+import ThirdPartyAuthAlert from './ThirdPartyAuthAlert';
 
 
 class LoginPage extends React.Component {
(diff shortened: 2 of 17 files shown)


src/data/constants.js
```diff
@@ -0,0 +1,8 @@
+// URL Paths
+export const LOGIN_PAGE = '/login';
+export const REGISTER_PAGE = '/register';
+export const RESET_PAGE = '/reset';
+export const DEFAULT_REDIRECT_URL = '/dashboard';
+
+// Constants
+export const SUPPORTED_ICON_CLASSES = ['apple', 'facebook', 'google', 'microsoft'];
```
waheedahmed, line 8: I believe normally we suggest `fontawesome` classes for `iconClass` while creating a provider configuration. Shouldn't we use the same values here? e.g.  fa-google-plus, fa-facebook, fa-linkedin.

waheedahmed (commented):
Couple of questions, otherwise looks good.

src/logistration/data/service.js
```diff
@@ -50,9 +50,10 @@ export async function login(creds) {
 export async function getThirdPartyAuthContext(urlParams) {
   const requestConfig = {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+    isPublic: true,
   };
 
-  const { data } = await getHttpClient()
+  const { data } = await getAuthenticatedHttpClient()
```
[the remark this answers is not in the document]
zainab-amir: This is necessary to associate the subsequent request with backend session so that we can get the "current provider" in the pipeline and user details.
If I use `getHttpClient`, TPA context request was unable to get this information as it always considered it an anonymous new user

src/data/constants.js
zainab-amir: In the code Fontawesome icons are used
https://github.com/edx/frontend-app-logistration/blob/cdd36bce1e9f4d8e1bdf39335c527bc18642679b/src/logistration/SocialAuthProviders.jsx#L41-L43

In order to use a string instead of `icon` object I had to pre-add these to the library and use this syntax
```
<FontAwesomeIcon icon=['fab', '<icon_string>'] />
```
The `SUPPORTED_ICON_CLASSES` is the list of icon strings accepted by `fontawesome`

src/logistration/data/service.js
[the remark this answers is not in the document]
waheedahmed: Make sense, thank you for clarifying.

waheedahmed (approved):

src/logistration/LoginPage.jsx
```diff
@@ -104,19 +104,29 @@ class LoginPage extends React.Component {
...
 
   render() {
+    const { currentProvider, finishAuthUrl, providers } = this.props.thirdPartyAuthContext;
+
     return (
       <>
         <RedirectLogistration
           success={this.props.loginResult.success}
           redirectUrl={this.props.loginResult.redirectUrl}
+          finishAuthUrl={finishAuthUrl}
```
adeelehsan, line 114: I think redirection won't happen for finishAuthUrl because it depends upon the success value and success value would be always false in your case.

src/logistration/RedirectLogistration.jsx
```diff
@@ -1,22 +1,33 @@
...
 function RedirectLogistration(props) {
-  const { success, redirectUrl } = props;
+  const { finishAuthUrl, success, redirectUrl } = props;
+
   if (success) {
-    window.location.href = redirectUrl;
+    // If we're in a third party auth pipeline, we must complete the pipeline
+    // once user has successfully logged in. Otherwise, redirect to the specified redirect url.
+    if (finishAuthUrl) {
+      window.location.href = getConfig().LMS_BASE_URL + finishAuthUrl;
```
adeelehsan, line 13: I think it should be independent of success. Because success would always false in your case because success is being passed from loginResult props.

zainab-amir: So basically, if the social account is associated with edX account, the pipeline will initiate the redirection to dashboard and we will never return back to MFE. 
Another scenario where above redirection is needed is when user doesn't have an edX account associated with social auth. In this case we come back to MFE in the middle of the running pipeline. We ask user to enter the usual email and password to his account and press "sign in". Once user clicks sign in `login_ajax` is called and we get loginResult back if the auth was successful. Now in this case where we have a pipeline running and login is successful, we want to redirect back to pipeline i.e `finishAuthUrl`.
If we don't send it to `finishAuthUrl` it will not be able to link social auth with edX account

adeelehsan (approved):

PR #28, opened 2020-11-15
Hide header menu links and removed footer.

uzairr:
VAN-100
VAN-101

`LOGISTRATION_MINIMAL_HEADER` env variable will be used in https://github.com/edx/frontend-component-header/pull/75 to hide the header links.


waheedahmed (approved):

adeelkhan (approved):
Just need to know how LOGISTRATION_MINIMAL_HEADER would play here. Iam looking at https://github.com/edx/frontend-component-header/pull/75/files to get the context. I think it would be used in future.

PR #31, opened 2020-11-19
Adds custom fields support to registration form.

adeelkhan:
VAN-19


src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -49,6 +50,233 @@ describe('./RegistrationPage.js', () => {
...
+            label: 'I agree to the Your Platform Name Here <a href="/honor" rel="noopener" target="_blank">Honor Code</a>',
+            name: 'honor_code',
+            type: 'checkbox',
+            errorMessages: {
+              required: 'You must agree to the Your Platform Name Here Honor Code',
+            },
+          }],
+        },
+      },
+    });
```
zainab-amir, line 75: spread operator should be called on `initialState` and not on the store (it is a MockStore object)

```suggestion
    store = mockStore({
      ...initialState,
      logistration: {
        ...initialState.logistration,
        registrationError: null,
        formData: {
          fields: [{
            label: 'I agree to the Your Platform Name Here <a href="/honor" rel="noopener" target="_blank">Honor Code</a>',
            name: 'honor_code',
            type: 'checkbox',
            errorMessages: {
              required: 'You must agree to the Your Platform Name Here Honor Code',
            },
          }],
        },
      },
    });
```

zainab-amir: update is required in all tests

src/common/Link.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { Hyperlink } from '@edx/paragon';
+
+const Link = (args) => (
+  <>
+    {args.beforeLink}
+    <Hyperlink destination={args.link}>
+      {args.linkText}
+    </Hyperlink>
+    {args.afterLink}
+  </>
```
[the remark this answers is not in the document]
adeelkhan: Two things for this;
1) Its already being used inside wrappers. https://github.com/edx/frontend-app-logistration/blob/master/src/logistration/ConfirmationAlert.jsx#L10. Secondly might work in above case, but adding multiple nodes under ValidationFormGroup even under <></> was causing error validations to fail. We might need to rename component that give better semantic meaning to whats happening.
2) In either case we need to factor out this code as its being used at multiple places. One is LoginFailure other is in Registration.

src/logistration/data/reducers.js
```diff
@@ -55,6 +57,19 @@ const reducer = (state = defaultState, action) => {
...
+      };
+    case REGISTER_FORM.SUCCESS:
+      return {
+        ...state,
+        formData: action.payload.formData,
+      };
+    case REGISTER_FORM.FAILURE:
+      return {
+        ...state,
+      };
```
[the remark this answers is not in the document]
adeelkhan: Hm. would remove it.

src/logistration/tests/RegistrationPage.test.jsx
adeelkhan: The rest i would refactor it

src/logistration/RegistrationPage.jsx
```diff
@@ -7,12 +7,29 @@ import {
...
+import { registerNewUser, fetchRegistrationForm } from './data/actions';
 import { registrationRequestSelector } from './data/selectors';
 import RedirectLogistration from './RedirectLogistration';
 import RegistrationFailure from './RegistrationFailure';
+import { processLink } from '../data/utils/dataUtils';
+import Link from '../common/Link';
+
+
+const REGISTRATION_VALIDITY_MAP = {};
+const REGISTRATION_EXTRA_FIELDS = [
```
waheedahmed, line 21: Move these new constants into https://github.com/edx/frontend-app-logistration/blob/3d0e5bef9f529ea38f2127a2ce725244167f9e45/src/data/constants.js#L1-L8

src/common/Link.jsx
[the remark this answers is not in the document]
waheedahmed: I second @zainab-amir's suggestion here.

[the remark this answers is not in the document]
zainab-amir: nit: I feel like the case in ConfirmationAlert is different from the case here (there we need to inject it inside FormatMessage and it makes the code cleaner).
Here we are doing the same thing in and outside of <Link> component. The messages `beforeLink` and `afterLink` are used as it is and we are not doing any additional processing on it inside <Link>. In fact this makes the use of Link in LoginFailure.jsx more complicated than using the HyperLink directly would have. I would probably get an other opinion on it.

src/logistration/RegistrationPage.jsx
```diff
@@ -127,21 +236,97 @@ class RegistrationPage extends React.Component {
...
   }
 
-  renderCountryList() {
-    const locale = getLocale();
-    let items = [{ value: '', label: 'Country or Region of Residence (required)' }];
-    items = items.concat(getCountryList(locale).map(({ code, name }) => ({ value: code, label: name })));
-    return items;
+  addExtraFields() {
+    const fields = this.props.formData.fields.map((value) => {
+      const field = value;
```
waheedahmed, line 263: I believe no need to define an extra variable here, you could simply do this:
```suggestion
    const fields = this.props.formData.fields.map((field) => {
```

src/logistration/RegistrationPage.jsx
```diff
@@ -127,21 +236,97 @@ class RegistrationPage extends React.Component {
...
+      if (field.required === true && REGISTRATION_EXTRA_FIELDS.includes(field.name)) {
+        const stateVar = camelCase(field.name);
```
waheedahmed, line 265: Instead of converting `field.name` in camel case at multiple places, we could simply convert the whole fields object into camel case using `import { camelCaseObject } from '@edx/frontend-platform';` at a single place.

src/logistration/LoginFailure.jsx
```diff
@@ -32,12 +15,11 @@ const LoginFailureMessage = (props) => {
       const [beforeLink, link, linkText, afterLink] = matches;
       return (
         <li key={error}>
-          <Link // eslint-disable-line  jsx-a11y/anchor-is-valid
-            beforeLink={beforeLink}
-            link={link}
-            linkText={linkText}
-            afterLink={afterLink}
-          />
+          <>
```
zainab-amir, line 18: nit: We don't need this

src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -57,13 +265,40 @@ describe('./RegistrationPage.js', () => {
   it('should match url after redirection', () => {
     const dasboardUrl = 'http://test.com/testing-dashboard/';
     store = mockStore({
-      ...store,
+      ...initialState,
       logistration: {
-        ...store.logistration,
+        ...initialState.logistration,
         registrationResult: {
+          ...initialState.logistration.registrationResult,
```
zainab-amir, line 272: nit: we don't need it since we are updating the `registrationResult` completely

src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -100,11 +331,37 @@ describe('./RegistrationPage.js', () => {
...
+              required: 'Enter your mailing address.',
+            },
+          }],
+        },
       },
     });
 
     const tree = renderer.create(reduxWrapper(<IntlRegistrationPage {...props} />)).toJSON();
     expect(tree).toMatchSnapshot();
+    windowSpy.mockClear();
```
zainab-amir, line 365: Can we move these to `afterEach`?

zainab-amir (approved):

src/logistration/RegistrationPage.jsx
adeelkhan: @waheedahmed camelCaseObject converts keys into camelCase. Here i need to convert the value. So i had to use the camelCase function directly.

Screen Shot 2020-11-27 at 1 44 30 PM

src/common/Link.jsx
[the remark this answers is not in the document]
adeelkhan: done

src/logistration/tests/RegistrationPage.test.jsx
adeelkhan: done

src/logistration/RegistrationPage.jsx
adeelkhan: done

src/logistration/RegistrationPage.jsx
adeelkhan: done

src/logistration/tests/RegistrationPage.test.jsx
adeelkhan: done

src/logistration/LoginFailure.jsx
adeelkhan: done

src/logistration/RegistrationPage.jsx
waheedahmed: Got it, thanks!

src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -120,6 +357,16 @@ describe('./RegistrationPage.js', () => {
           ...initialState.logistration.thirdPartyAuthContext,
           secondaryProviders: [secondaryProviders],
         },
+        formData: {
```
waheedahmed, line 360: I believe this is required in each test case, can we set this in `initialState` so that we don't need to repeat this in. each test?
https://github.com/edx/frontend-app-logistration/blob/f0fad2b4cb308e3de372738cc395c0ed7e354a25/src/logistration/tests/RegistrationPage.test.jsx#L18-L28

waheedahmed (approved):
Looks good except for one question.

waheedahmed (approved):

src/logistration/data/service.js
```diff
@@ -66,3 +66,23 @@ export async function getThirdPartyAuthContext(urlParams) {
...
   };
 }
+
+export async function getRegistrationForm() {
+  const requestConfig = {
+    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+    isPublic: true,
+  };
+
+  const { data } = await getAuthenticatedHttpClient()
```
zainab-amir, line 76: can we use simple HTTPClient if the API is public?

src/logistration/RegistrationPage.jsx
```diff
@@ -64,10 +93,16 @@ class RegistrationPage extends React.Component {
...
-      honor_code: true,
-      country: this.state.country,
+      name: this.state.name,
     };
+
+    const fieldMap = { ...REGISTRATION_VALIDITY_MAP, ...REGISTRATION_OPTIONAL_MAP };
+    Object.keys(fieldMap).forEach((key) => {
+      const value = fieldMap[key];
+      if (value) {
+        payload[key] = this.state[camelCase(key)];
```
zainab-amir, line 103: why are we not getting the camel cased keys directly inside the service when we get the data?

src/logistration/data/service.js
adeelkhan: I dont know for sure... But how much difference it would make on performance.

src/logistration/RegistrationPage.jsx
adeelkhan: If the question is that we should be handling the object directly. The issue is camelCase here is needed in value rather than for keys. So had to be used directly.

zainab-amir: okay

src/logistration/data/service.js
zainab-amir: `getHttpClient()` is provided by @edx/frontend-platform/auth and can be used when the the API is public (as suggested in the header isPublic: True)

zainab-amir: I am not sure if using one over the other should have any performance impact

src/logistration/RegistrationPage.jsx
adeelkhan: While the above answer does hold for when we are not converting api data keys into camelCase, in this specific case, I am does using camelCase on key. The thing is that i have creating a mapping/dictionary ie fieldMap  which is used to filter out fields based on their name/id which are not in camelCase. I think we can improve on it. Making this mapping in camelCase we might remove at least this direct call all together.

zainab-amir: yeah

zainab-amir (approved):
A few changes are needed but are not critical for budbush and can be done later

PR #32, opened 2020-11-20
Add SSO to register page

zainab-amir:
- Added SSO to register page
- Added tests for SSO
- Updated few css classes

Note: This is missing the autofill register form functionality that populates user information received from social auth. It will be covered when we use formik for forms.

Screenshot 2020-11-20 at 1 16 51 PM

Screenshot 2020-11-19 at 5 29 44 PM

### DIFF
--- a/src/logistration/tests/RegistrationPage.test.jsx
+++ b/src/logistration/tests/RegistrationPage.test.jsx
@@ -3,6 +3,7 @@ import { Provider } from 'react-redux';
 import renderer from 'react-test-renderer';
 import configureStore from 'redux-mock-store';
 import { mount } from 'enzyme';
+import { getConfig } from '@edx/frontend-platform';
 import { IntlProvider, injectIntl, configure } from '@edx/frontend-platform/i18n';
 
 import RegistrationPage from '../RegistrationPage';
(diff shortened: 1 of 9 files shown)


src/logistration/tests/RegistrationPage.test.jsx
```diff
@@ -113,6 +142,75 @@ describe('./RegistrationPage.js', () => {
     expect(root.text().includes('Test University')).toBe(true);
   });
 
+  it('should match url after TPA redirection', () => {
+    const authCompleteUrl = '/auth/complete/google-oauth2/';
+    store = mockStore({
+      ...initialState,
+      logistration: {
+        ...initialState.logistration,
+        registrationResult: {
```
adeelehsan, line 151: default values should work. Don't need to define again.
Please remove other test cases as well.

adeelehsan (approved):
Overall looks good.  You can merge it after updating the tests.

src/logistration/tests/RegistrationPage.test.jsx
zainab-amir: I don't want to use the default values, I need the values to be `true`, that is why I have added it here in test

waheedahmed (approved):

PR #48, opened 2020-12-22
Add inactive user error message

zainab-amir:
#### Changes:
- Added `INACTIVE_USER` error check in LoginFailure component
- Removed inline style
- Replace links and headings with Alert.Link and Alert.Heading (https://edx.github.io/paragon/components/alert)
- Removed test from LoginPage as those don't contribute to coverage and should be added to LoginFailure tests

[VAN-217](https://openedx.atlassian.net/browse/VAN-217)

### DIFF
--- a/src/logistration/LoginFailure.jsx
+++ b/src/logistration/LoginFailure.jsx
@@ -5,17 +5,17 @@ import { Alert, Hyperlink } from '@edx/paragon';
 import PropTypes from 'prop-types';
 
 import { processLink } from '../data/utils/dataUtils';
-import { NON_COMPLIANT_PASSWORD_EXCEPTION } from './data/constants';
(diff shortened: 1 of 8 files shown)


src/logistration/LoginFailure.jsx
```diff
@@ -31,9 +31,38 @@ const LoginFailureMessage = (props) => {
...
+          <FormattedMessage
+            id="logistration.contact.support.link"
+            defaultMessage="contact {platformName} Support"
+            description="Link text used in inactive user error message to go to learner help center"
+            values={{ platformName: context.platformName }}
+          />
+        </Alert.Link>
+      );
+      errorList = (
+        <li key="inactive-user">
```
waheedahmed, line 47: nit: We could utilize `INACTIVE_USER` here.

waheedahmed (approved):
One minor nit otherwise 👍

PR #49, opened 2020-12-23
Rearrange optional fields

uzairr:
Rearrange optional fields to make them compliant with old monolith


waheedahmed (changes requested):
Instead of our custom styling use paragon's `Form.Row` https://edx.github.io/paragon/components/forms/.

uzairr:
@waheedahmed Form.Row is not working in our case, if we follow the form creation on the same lines described in the above shared paragon article then input drop-downs cant be created which is required in this case.

waheedahmed (approved):

PR #50, opened 2020-12-23
Fix form submission issues.

waheedahmed:
Fixed form submission issues on the first click and with optional fields check. Also fixed some warnings and did some refactoring as well.

VAN-225, VAN-206

### DIFF
--- a/src/logistration/RegistrationPage.jsx
+++ b/src/logistration/RegistrationPage.jsx
@@ -74,7 +74,7 @@ class RegistrationPage extends React.Component {
       usernameValid: false,
       passwordValid: false,
       countryValid: false,
-      honorCodeValid: false,
+      honorCodeValid: true,
       termsOfServiceValid: false,
       formValid: false,
       institutionLogin: false,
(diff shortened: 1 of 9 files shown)


src/logistration/RegistrationPage.jsx
```diff
@@ -353,34 +349,33 @@ class RegistrationPage extends React.Component {
...
+              label: item.name,
+            }));
+            props.options = options;
           }
+          return (
+            <ValidationFormGroup
+              for={field.name}
+              key={field.name}
+            >
+              <label htmlFor={field.name} className="h6 pt-3">{field.label} (optional)</label>
```
zainab-amir, line 372: do we need to mark (optional) as translation string?

waheedahmed: @zainab-amir yes, it should be. Let me do that.

waheedahmed: done

zainab-amir (approved):

PR #52, opened 2020-12-24
This patches fixes following;

adeelkhan:
*) Empty fields error messages
*) Removes onChange validations

VAN-204

### DIFF
--- a/src/logistration/RegistrationPage.jsx
+++ b/src/logistration/RegistrationPage.jsx
@@ -61,6 +61,7 @@ class RegistrationPage extends React.Component {
       confirmEmail: '',
       enableOptionalField: false,
       validationFieldName: '',
+      emptyFields: {},
       errors: {
         email: '',
         name: '',
@@ -96,8 +97,31 @@ class RegistrationPage extends React.Component {
       const { errors } = this.state;
       const errorMsg = nextProps.validations.validation_decisions[this.state.validationFieldName];
       errors[this.state.validationFieldName] = errorMsg;
-      this.setState({
-        errors,
+      const stateValidKey = `${camelCase(this.state.validationFieldName)}Valid`;
+      const stateValidValue = !errorMsg;
+
+      this.setState(({ [stateValidKey]: stateValidValue }), () => {
+        const {
+          emailValid,
+          nameValid,
+          usernameValid,
+          passwordValid,
+        } = this.state;
+
+        const validityMap = REGISTRATION_VALIDITY_MAP;
+        let extraFieldsValid = true;
+        Object.entries(validityMap).forEach(([key, value]) => {
+          if (value) {
+            const stateValid = `${camelCase(key)}Valid`;
+            extraFieldsValid = extraFieldsValid && this.state[stateValid];
+          }
+        });
+
+        const formValid = emailValid && nameValid && usernameValid && passwordValid && extraFieldsValid;
+        this.setState({
+          errors,
+          formValid,
+        });
       });
       return false;
     }
@@ -112,10 +136,10 @@ class RegistrationPage extends React.Component {
     e.preventDefault();
     const params = (new URL(document.location)).searchParams;
     const payload = {
-      email: this.state.email,
-      username: this.state.username,
-      password: this.state.password,
       name: this.state.name,
+      username: this.state.username,
+      email: this.state.email,
+      password: this.state.password,
     };
 
     const fieldMap = { ...REGISTRATION_VALIDITY_MAP, ...REGISTRATION_OPTIONAL_MAP };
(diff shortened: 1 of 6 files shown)


src/logistration/RegistrationPage.jsx
```diff
@@ -192,24 +215,44 @@ class RegistrationPage extends React.Component {
 
     switch (inputName) {
       case 'email':
-        emailValid = value.match(/^([\w.%+-]+)@([\w-]+\.)+([\w]{2,})$/i);
-        errors.email = emailValid ? '' : this.intl.formatMessage(messages['logistration.email.validation.message']);
+        emailValid = value.length >= 1;
+        if (!emailValid) {
+          emptyFields.email = [{ user_message: this.intl.formatMessage(messages['logistration.email.validation.message']) }];
+        } else {
+          emptyFields.email = [{ user_message: '' }];
```
waheedahmed, line 222: Can we handle all these errors using a utility function like the following?
```
  generateUserMessage = (isFieldValid, messageID) => {
    return [{ user_message: isFieldValid ? '' : this.intl.formatMessage(messages[messageID] }];
  }
```
```suggestion
        emptyFields.email = generateUserMessage(emailValid, 'logistration.email.validation.message']);
```

src/logistration/RegistrationPage.jsx
```diff
@@ -96,8 +97,31 @@ class RegistrationPage extends React.Component {
       const { errors } = this.state;
       const errorMsg = nextProps.validations.validation_decisions[this.state.validationFieldName];
       errors[this.state.validationFieldName] = errorMsg;
-      this.setState({
-        errors,
+      const stateValidVar = `${camelCase(this.state.validationFieldName)}Valid`;
```
waheedahmed, line 100: nit: `stateValidKey`

src/logistration/RegistrationPage.jsx
```diff
@@ -96,8 +97,31 @@ class RegistrationPage extends React.Component {
...
+      const stateValidValue = !errorMsg;
+
+      this.setState(() => ({ [stateValidVar]: stateValidValue }), () => {
```
waheedahmed, line 103: nit:
```suggestion
      this.setState(({ [stateValidVar]: stateValidValue }), () => {
```

waheedahmed (changes requested):
Also, React suggests that we should avoid extensive operations inside `shouldComponentUpdate()`, which can harm the performance. Is there any other way of doing what we are currently doing inside it?

src/logistration/RegistrationPage.jsx
adeelkhan: Done

src/logistration/RegistrationPage.jsx
adeelkhan: Done

src/logistration/RegistrationPage.jsx
adeelkhan: Done

src/logistration/RegistrationPage.jsx
```diff
@@ -192,24 +215,44 @@ class RegistrationPage extends React.Component {
...
+        if (!usernameValid) {
+          emptyFields.username = [{ user_message: this.intl.formatMessage(messages['logistration.username.validation.message']) }];
+        } else {
+          emptyFields.username = [{ user_message: '' }];
+        }
         break;
       case 'password':
-        passwordValid = !!(value.length >= 8 && value.match(/\d+/g));
-        errors.password = passwordValid ? '' : this.intl.formatMessage(messages['logistration.register.page.password.validation.message']);
+        passwordValid = !!(value.length >= 1);
```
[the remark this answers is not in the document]
adeelkhan: Done

adeelkhan (commented):
> Also, React suggests that we should avoid extensive operations inside `shouldComponentUpdate()`, which can harm the performance. Is there any other way of doing what we are currently doing inside it?

Two things here; 
1) The working here are not too expensive in nature. Just setting state and in async testing form validation.
2) I have to update state before render, and this is one of way. I don't know for any other solution for now. We can fix forward.

src/logistration/_style.scss
```diff
@@ -218,3 +218,7 @@ $apple-focus-black: $apple-black;
 .opt-year-field{
     padding-left: 15px;
 }
+
+.invalid-feedback {
+  color: #b20610
```
waheedahmed, line 223: Use theme variables `$danger` or `$red`.

waheedahmed (approved):
One minor change otherwise 👍

src/logistration/_style.scss
adeelkhan: Done

PR #54, opened 2020-12-28
add enterprise login link

uzairr:
Add enterprise login link to MFE so that enterprise user would be redirected to enterprise login page.

VAN-214

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -5,13 +5,15 @@ export const RESET_PAGE = '/reset';
 export const DEFAULT_REDIRECT_URL = '/dashboard';
 export const PASSWORD_RESET_CONFIRM = '/password_reset_confirm/:token/';
 export const PAGE_NOT_FOUND = '/notfound';
+export const ENTERPRISE_LOGIN_URL = '/enterprise/login';
+
+// Constants
+export const SUPPORTED_ICON_CLASSES = ['apple', 'facebook', 'google', 'microsoft'];
 
 // Stateful Submit Button States
 export const DEFAULT_STATE = 'default';
 export const PENDING_STATE = 'pending';
 
-// Constants
-export const SUPPORTED_ICON_CLASSES = ['apple', 'facebook', 'google', 'microsoft'];
 export const REGISTRATION_VALIDITY_MAP = {};
 export const REGISTRATION_OPTIONAL_MAP = {};
 export const REGISTRATION_EXTRA_FIELDS = [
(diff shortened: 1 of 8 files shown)


src/logistration/LoginHelpLinks.jsx
```diff
@@ -61,6 +75,9 @@ const LoginHelpLinks = (props) => {
           default: <></>,
         }}
       />
+      <button type="button" className="mt-1 field-link position-relative" style={{ top: '-10px' }}>
```
waheedahmed, line 78: Remove this unnecessary button element around the link and use `mt-0` class instead of setting inline styles.

waheedahmed: We also need to add this in `LoginPage.jsx` instead of this component otherwise, it will add this link on the forgot password page as well.

waheedahmed (changes requested):

src/data/constants.js
```diff
@@ -3,6 +3,7 @@ export const LOGIN_PAGE = '/login';
 export const REGISTER_PAGE = '/register';
 export const RESET_PAGE = '/reset';
 export const DEFAULT_REDIRECT_URL = '/dashboard';
+export const ENTERPRISE_LOGIN_URL = '/enterprise/login';
```
waheedahmed, line 6: This is an edX specific endpoint, we should not hard code it here and instead pull it from remote config.

waheedahmed (changes requested):

src/logistration/messages.jsx
```diff
@@ -191,6 +191,11 @@ const messages = defineMessages({
     defaultMessage: '(optional)',
     description: 'Text that appears with optional field labels',
   },
+  'logistration.enterprise.login.help.message': {
+    id: 'logistration.enterprise.login.help.message',
+    defaultMessage: 'Sign in with your company or school',
+    description: 'Text that helps enterprise user to login',
```
waheedahmed, line 197: `Company or school login link text.`

src/logistration/messages.jsx
```diff
@@ -191,6 +191,11 @@ const messages = defineMessages({
     defaultMessage: '(optional)',
     description: 'Text that appears with optional field labels',
   },
+  'logistration.enterprise.login.help.message': {
```
waheedahmed, line 194: `logistration.enterprise.login.link.text`

waheedahmed (approved):

PR #55, opened 2020-12-28
re-align loader

uzairr:
Re-align Loader

VAN-111


waheedahmed (approved):

waheedahmed (approved):
Please move inline styles into the styles file, otherwise looks good 👍

waheedahmed (approved):

PR #56, opened 2020-12-28
add email placeholder

uzairr:
Add placeholder to email field in registration form.

VAN-155


src/logistration/RegistrationPage.jsx
```diff
@@ -507,7 +507,7 @@ class RegistrationPage extends React.Component {
                 name="email"
                 id="email"
                 type="email"
-                placeholder=""
+                placeholder="Email"
```
waheedahmed, line 510: I guess it should be `username@domain.com`.

waheedahmed (approved):

PR #58, opened 2020-12-31
Add error handling for 500 status code

zainab-amir:
This handles the cases where some internal server error occurs and no error message is shown to users.
Screenshot 2020-12-31 at 2 29 46 PM

[VAN-95](https://openedx.atlassian.net/browse/VAN-95)


waheedahmed (commented):
Don't we need to handle this on all pages?

waheedahmed (approved):

PR #59, opened 2020-12-31
update regex for email on forgot password page

zainab-amir:
Updated the regex to be same as FE 

Screenshot 2020-12-31 at 6 06 10 PM

[VAN-211](https://openedx.atlassian.net/browse/VAN-211)

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -4,13 +4,12 @@ export const REGISTER_PAGE = '/register';
 export const RESET_PAGE = '/reset';
 export const DEFAULT_REDIRECT_URL = '/dashboard';
 
-// Constants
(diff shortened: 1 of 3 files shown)


src/data/constants.js
```diff
@@ -25,3 +24,8 @@ export const REGISTRATION_EXTRA_FIELDS = [
   'city',
   'country',
 ];
+
+export const VALID_EMAIL_FORMAT = '(^[-!#$%&\'*+/=?^_`{}|~0-9A-Z]+(\\.[-!#$%&\'*+/=?^_`{}|~0-9A-Z]+)*'
```
waheedahmed, line 28: nit: `VALID_EMAIL_REGEX`

waheedahmed (approved):
One minor suggestion otherwise 👍

PR #61, opened 2020-12-31
Fix password reset confirmation redirection.

waheedahmed:
Fixed password reset redirection to dashboard page if user is already logged in.

VAN-234


src/index.jsx
```diff
@@ -25,21 +27,21 @@ registerIcons();
...
-      </LoggedInRedirect>
+      <HeaderLayout>
+        <Switch>
+          <Route exact path="/">
+            <Redirect to={PAGE_NOT_FOUND} />
+          </Route>
+          <UnAuthOnlyRoute exact path={LOGIN_PAGE} component={LoginPage} />
+          <UnAuthOnlyRoute exact path={REGISTER_PAGE} component={RegistrationPage} />
+          <UnAuthOnlyRoute exact path={RESET_PAGE} component={ForgotPasswordPage} />
+          <Route exact path="/password_reset_confirm/:token/" component={ResetPasswordPage} />
```
zainab-amir, line 38: nit: can we move this path to constants as well?

zainab-amir (approved):
Looks good. Just added one suggestion

src/index.jsx
waheedahmed: sure.

PR #65, opened 2021-01-04
fix spinner position in safari

uzairr:
Align Spinner position to center across the browsers.

VAN-111


waheedahmed (changes requested):
Youn could achieve this without any custom CSS using already available bootstrap classes:
```
const Spinner = () => (
  <div className="container position-absolute h-100">
    <div className="d-flex justify-content-center align-items-center h-100">
      <ParagonSpinner animation="border" variant="primary" />
    </div>
  </div>
);
```

waheedahmed (approved):

PR #66, opened 2021-01-04
populate registration fields

zainab-amir:
When registering via SSO, populate registration fields with user data acquired from the social auth pipeline.

[VAN-238](https://openedx.atlassian.net/browse/VAN-238)


adeelkhan (approved):
Squash commits

PR #69, opened 2021-01-05
Make logistration pages responsive.

waheedahmed:
Made logistration pages response for smaller devices using paragon components and bootstrap classes.

VAN-222

**Sandbox:** https://logistration-van-222.sandbox.edx.org/login


src/logistration/InstitutionLogistration.jsx
```diff
@@ -34,23 +34,22 @@ const InstitutionLogistration = props => {
...
+          <h3 className="mt-5 mb-4 font-weight-normal">
             {headingTitle}
-          </p>
-          <div style={{ fontSize: '16px' }}>
-            <p
-              className="mb-2"
-              style={{ fontSize: '16px' }}
-            >
+          </h3>
+          <div className="">
```
zainab-amir, line 42: nit: remove empty className

src/logistration/_style.scss
```diff
@@ -21,8 +21,6 @@ $apple-focus-black: $apple-black;
 
 .register-container {
   @extend .login-container;
-
-  width: 30rem;
```
zainab-amir, line 25: if login-container is not used anywhere else can we remove it and just add the values directly to register-container?

zainab-amir (approved):
Looks good 💯 Just a couple of nits

src/logistration/_style.scss
waheedahmed: @zainab-amir Actually both are not being used, I'm gonna remove both.

PR #74, opened 2021-01-08
fix missing branding issues

uzairr:
fix missing issues related to branding.


zainab-amir (approved):
Looks good just one thing,
if `text-center` and `submit` are custom css and are not being used anywhere, can we remove those from scss file?

uzairr:
@zainab-amir `text-center` is from bootstrap library and in case of `submit`, we need to verify it whether it has a ripple effect or not.

PR #79, opened 2021-01-18
Handle 500 api response

uzairr:
Screen Shot 2021-01-27 at 7 26 31 PM

VAN-281

### DIFF
--- a/src/register/RegistrationFailure.jsx
+++ b/src/register/RegistrationFailure.jsx
@@ -1,47 +1,55 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Alert } from '@edx/paragon';
-
-const hasNoErrors = (userErrors) => (
-  userErrors.every((errorList) => (!errorList[0]))
-);
+import { INTERNAL_SERVER_ERROR } from '../login/data/constants';
+import messages from './messages';
 
 const RegistrationFailureMessage = (props) => {
   const errorMessage = props.errors;
+  const { errorCode } = props.errors;
   const userErrors = [];
 
   useEffect(() => {
     window.scrollTo({ left: 0, top: 0, behavior: 'smooth' });
   }, [props.submitCount]);
 
-  Object.keys(errorMessage).forEach((key) => {
-    const errors = errorMessage[key];
-    const errorList = errors.map((error) => (
-      (error.user_message) ? (
-        <li key={error} className="text-left">
-          {error.user_message}
+  let serverError;
+  switch (errorCode) {
+    case INTERNAL_SERVER_ERROR:
+      serverError = (
+        <li key={INTERNAL_SERVER_ERROR} className="text-left">
+          {props.intl.formatMessage(messages['registration.request.server.error'])}
         </li>
-      ) : null
-    ));
-    userErrors.push(errorList);
-  });
+      );
+     userErrors.push(serverError);
+     break;
+
+     default:
+        Object.keys(errorMessage).forEach((key) => {
+          const errors = errorMessage[key];
+          const errorList = errors.map((error) => (
+            (error.user_message) ? (
+              <li key={error} className="text-left">
+                {error.user_message}
+              </li>
+            ) : null
+          ));
+          userErrors.push(errorList);
+        });
+  }
 
   return (
-    hasNoErrors(userErrors) ? null : (
+    !userErrors.length ? null : (
       <Alert variant="danger">
         <Alert.Heading>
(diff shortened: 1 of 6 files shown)


zainab-amir (commented):
Can we use the approach similar to the one used in `LoginFailure.jsx`
Since eventually we are going to move to error codes instead of parsing errors coming from backend, it would be better to have an option that is flexible to incorporate those changes later.

src/logistration/RegistrationFailure.jsx
```diff
@@ -2,48 +2,61 @@ import React, { useEffect } from 'react';
...
+        id="logistration.registration.request.failure.header.description.message"
+        defaultMessage="We couldn't create your account."
+        description="error message when registration failure."
+      />
+    </Alert.Heading>
+    <div>
+      <ul>
+        {userErrors}
+      </ul>
+    </div>
```
zainab-amir, line 20: do we need div around `<ul>`?

src/logistration/RegistrationFailure.jsx
```diff
@@ -2,48 +2,61 @@ import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { Alert } from '@edx/paragon';
+import { INTERNAL_SERVER_ERROR } from './data/constants';
 
-const hasNoErrors = (userErrors) => (
-  userErrors.every((errorList) => (!errorList[0]))
+const showAlert = (userErrors) => (
+  <Alert variant="danger">
+    <Alert.Heading className="text-danger">
```
zainab-amir, line 9: We don't need the class `text-danger` as UX has suggested we use the same color heading and body text.

src/logistration/RegistrationFailure.jsx
```diff
@@ -2,48 +2,61 @@ import React, { useEffect } from 'react';
...
+      <FormattedMessage
+        id="logistration.registration.request.failure.header.description.message"
+        defaultMessage="We couldn't create your account."
+        description="error message when registration failure."
+      />
```
zainab-amir, line 14: We should add these string to messages.jsx 
nit: Also the id is too long, `registration.request.failure.header` is sufficient

src/logistration/RegistrationFailure.jsx
```diff
@@ -2,48 +2,61 @@ import React, { useEffect } from 'react';
...
-          />
-        </Alert.Heading>
-        <div>
-          <ul>
-            {userErrors}
-          </ul>
-        </div>
-      </Alert>
-    )
+    !userErrors.length ? null : showAlert(userErrors)
```
zainab-amir, line 59: I think we keep the `<Alert>` elements in the return statement instead of creating a wrapper around it. Something like this can be done:
https://github.com/edx/frontend-app-logistration/blob/ad2ee9e013781618d9df01bf4e28217441562992/src/login/AccountActivationMessage.jsx#L50-L55

src/logistration/RegistrationFailure.jsx
```diff
@@ -2,48 +2,61 @@ import React, { useEffect } from 'react';
...
-    const errorList = errors.map((error) => (
-      (error.user_message) ? (
-        <li key={error} className="text-left">
-          {error.user_message}
+  let serverError;
+  switch (errorCode) {
+    case INTERNAL_SERVER_ERROR:
+      serverError = (
+        <li key={INTERNAL_SERVER_ERROR} className="text-left">
+          An error has occurred. Try refreshing the page, or check your Internet connection.
```
waheedahmed, line 38: Wrap this with `FormattedMessage` or move this into `messages.jsx`.

src/logistration/RegistrationFailure.jsx
waheedahmed: +1 on this.

src/register/RegistrationFailure.jsx
```diff
@@ -1,47 +1,55 @@
...
+        Object.keys(errorMessage).forEach((key) => {
+          const errors = errorMessage[key];
+          const errorList = errors.map((error) => (
+            (error.user_message) ? (
+              <li key={error} className="text-left">
+                {error.user_message}
+              </li>
+            ) : null
+          ));
+          userErrors.push(errorList);
```
waheedahmed, line 38: Why we are pushing an array into another array here? I guess we should `userErrors.concat(errorList);` instead of push.

waheedahmed (commented):
Just one question otherwise 👍

src/register/RegistrationFailure.jsx
uzairr: @waheedahmed  `userErrors` must contain a list of `li` elements in order to render them on the top of the page.Concat doesn't work here because it will concat all the li's in each successive iteration.

waheedahmed (approved):

PR #80, opened 2021-01-18
Change login and registration API endpoints.

waheedahmed:
Changed login/registration endpoints to match monolith FE. Also fixed a warning and removed an extra call to `login_refresh`.

VAN-107
VAN-12

### DIFF
--- a/src/reset-password/data/service.js
+++ b/src/reset-password/data/service.js
@@ -27,15 +27,14 @@ export async function resetPassword(payload, token, queryParams) {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
     isPublic: true,
   };
+  const url = new URL(`${getConfig().LMS_BASE_URL}/password/reset/${token}/`);
 
-  let path = `password/reset/${token}/?track=pwreset`;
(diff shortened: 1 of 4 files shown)


adeelkhan (approved):

src/reset-password/data/service.js
```diff
@@ -27,15 +27,14 @@ export async function resetPassword(payload, token, queryParams) {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
     isPublic: true,
   };
+  const url = new URL(`${getConfig().LMS_BASE_URL}/password/reset/${token}/`);
```
adeelkhan, line 30: I dont think we use this query param, Just for parity sake. We can remove it in future if we need. 
```suggestion
  const url = new URL(`${getConfig().LMS_BASE_URL}/password/reset/${token}/?track=pwreset`);
```

waheedahmed: Since this is already part of the password reset URL, we don't need to set it on the API endpoint as well. This query param is just used to track the password reset page in google analytics.

PR #81, opened 2021-01-18
Logistration renamed to authn

adeelehsan:


waheedahmed:
@adeelehsan Do we still need this PR? Otherwise could you please close this out?

PR #82, opened 2021-01-19
Add account activation messages

zainab-amir:
Show messages based on the query paran recieved during the redirect after
account activation link is clicked

VAN-304

### DIFF
--- a/src/logistration/AccountActivationMessage.jsx
+++ b/src/logistration/AccountActivationMessage.jsx
@@ -1,0 +1,63 @@
+import React from 'react';
+
+import { getConfig } from '@edx/frontend-platform';
+import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Alert } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+import { ACCOUNT_ACTIVATION_MESSAGE } from './data/constants';
+import messages from './messages';
+
+const AccountActivationMessage = (props) => {
+  const { intl, messageType } = props;
+  const variant = messageType === ACCOUNT_ACTIVATION_MESSAGE.ERROR ? 'danger' : messageType;
+
+  let activationMessage;
+  let heading;
+
+  switch (messageType) {
+    case ACCOUNT_ACTIVATION_MESSAGE.SUCCESS: {
+      heading = intl.formatMessage(messages['authn.account.activation.success.message.title']);
+      activationMessage = intl.formatMessage(messages['authn.account.activation.success.message']);
+      break;
+    }
+    case ACCOUNT_ACTIVATION_MESSAGE.INFO: {
+      activationMessage = intl.formatMessage(messages['authn.account.already.activated.message']);
+      break;
+    }
+    case ACCOUNT_ACTIVATION_MESSAGE.ERROR: {
+      const supportLink = (
+        <Alert.Link href={getConfig().ACTIVATION_EMAIL_SUPPORT_LINK}>
+          {intl.formatMessage(messages['authn.account.activation.support.link'])}
+        </Alert.Link>
+      );
+
+      heading = intl.formatMessage(messages['authn.account.activation.error.message.title']);
+      activationMessage = (
+        <FormattedMessage
+          id="authn.account.activation.error.message"
+          defaultMessage="Something went wrong, please {supportLink} to resolve this issue."
+          description="Account activation error message"
+          values={{ supportLink }}
+        />
+      );
+      break;
+    }
+    default:
+      break;
+  }
+
+  return activationMessage ? (
+    <Alert id="account-activation-message" variant={variant}>
+      {heading && <Alert.Heading>{heading}</Alert.Heading>}
+      {activationMessage}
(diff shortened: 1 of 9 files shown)


src/logistration/LoginPage.jsx
```diff
@@ -147,6 +148,9 @@ class LoginPage extends React.Component {
     } = this.props;
     const { currentProvider, providers, secondaryProviders } = this.props.thirdPartyAuthContext;
 
+    const params = (new URL(window.location.href)).searchParams;
+    const activationMsgType = params.get('next') ? params.get('next').split('activate_account=')[1] : null;
```
waheedahmed, line 152: Why we are checking for `next` query param here? Can't we simply get the `activate_account` query param like this?
```suggestion
    const activationMsgType = params.get('activate_account');
```

src/logistration/tests/AccountActivationMessage.test.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
waheedahmed, line 1: Instead of snapshot testing, can we simply test each case in `LoginPage.test.jsx` like you are doing for `info`?

zainab-amir: should I just remove these tests?

waheedahmed: Yes, those tests will cover each scenario.

src/logistration/LoginPage.jsx
zainab-amir: No. Because when user redirects to dashboard and is not authenticated, the query param next is added by the @login_required decorator and this just gets appended to the url 
```
http://localhost:1999/login?next=/dashboard?activate_account=error
```
if it was added with an `&` we could have fetched it directly
```
http://localhost:1999/login?next=/dashboard&activate_account=error
```

src/logistration/AccountActivationMessage.jsx
```diff
@@ -0,0 +1,77 @@
...
+import { Alert } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+import { ACCOUNT_ACTIVATION_MESSAGE } from './data/constants';
+
+const AccountActivationMessage = (props) => {
+  const { messageType } = props;
+
+  const supportLink = (
+    <Alert.Link href="https://support.edx.org/hc/en-us/articles/227340127-Why-haven-t-I-received-my-activation-email-">
```
zainab-amir, line 13: @waheedahmed I almost forgot. What is the best way to add these urls? I don't want it hardcoded

waheedahmed: Thorugh an environment variable using remote configs.

src/logistration/tests/AccountActivationMessage.test.jsx
waheedahmed: Or, if you want to test it here then instead of snapshots verify the actual message for each case.

src/logistration/AccountActivationMessage.jsx
waheedahmed: And yes, we should not hardcode edX specific things here.

src/logistration/LoginPage.jsx
waheedahmed: Got it, but I guess we should avoid that extra redirect. Check my comment on the other PR.

src/logistration/tests/AccountActivationMessage.test.jsx
zainab-amir: I have kept the file and verified the actual messages

src/logistration/AccountActivationMessage.jsx
zainab-amir: I have used env variable for this. Will create PR for edx-internal

src/logistration/AccountActivationMessage.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+      heading = intl.formatMessage(messages['authn.account.activation.error.message.title']);
+      activationMessage = (
+        <FormattedMessage
+          id="authn.account.activation.error.message"
+          defaultMessage="Something went wrong, please {supportLink} to resolve this issue."
+          description="Account activation error message"
+          values={{ supportLink }}
+        />
+      );
```
waheedahmed, line 43: You can move this into the messages as well and can pass the values to `intl.formatMessage()` like this:
```
      activationMessage = intl.formatMessage(
          messages['authn.account.activation.error.message'],
          { supportLink }
      );
```

waheedahmed (approved):
One comment otherwise 👍

src/logistration/AccountActivationMessage.jsx
zainab-amir: I didn't know this. Thank you, will do

zainab-amir: I don't think this will work in this case because `supportLink` is an object of `<Alert.Link>`

adeelkhan (approved):

PR #84, opened 2021-01-21
Handle 500 server error on password reset page

uzairr:
Screen Shot 2021-01-27 at 7 42 59 PM

VAN-281

### DIFF
--- a/src/reset-password/ResetPasswordPage.jsx
+++ b/src/reset-password/ResetPasswordPage.jsx
@@ -16,6 +16,8 @@ import { validatePassword } from './data/service';
 import InvalidTokenMessage from './InvalidToken';
 import ResetSuccessMessage from './ResetSuccess';
 import Spinner from './Spinner';
+import APIFailureMessage from '../common-components/APIFailureMessage';
+import { INTERNAL_SERVER_ERROR } from '../data/constants';
 
 const ResetPasswordPage = (props) => {
   const { intl } = props;
@@ -78,8 +80,16 @@ const ResetPasswordPage = (props) => {
       props.validateToken(token);
       return <Spinner />;
     }
-  } else if (props.token_status === 'invalid') {
+  } else if (props.token_status === 'invalid' && props.errorCode !== INTERNAL_SERVER_ERROR) {
     return <InvalidTokenMessage />;
+  } else if (props.token_status === 'invalid' && props.errorCode === INTERNAL_SERVER_ERROR) {
+    return (
+      <div className="d-flex justify-content-center m-4">
+        <div className="d-flex flex-column mw-500">
+          <APIFailureMessage header={intl.formatMessage(messages['reset.password.token.validation.sever.error'])} />
+        </div>
+      </div>
+    );
   } else if (props.status === 'success') {
     return <ResetSuccessMessage />;
   } else {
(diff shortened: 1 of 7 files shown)


waheedahmed:
Please add screenshots.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -65,7 +67,9 @@ const ResetPasswordPage = (props) => {
       props.resetPassword(formPayload, props.token, params);
     }
   };
-
+  if (props.errorCode === INTERNAL_SERVER_ERROR) {
+    return (<APIFailureMessage />);
```
waheedahmed, line 71: Can we use the same approach used at [L88](https://github.com/edx/frontend-app-authn/pull/84/files#diff-62ac94b242d30a85a60109539e352737262794426cba6505ec68c8459004e65bR88)?

src/reset-password/APIFailureMessage.jsx
```diff
@@ -0,0 +1,28 @@
...
+      <div className="text-center mw-420">
+        <Alert variant="danger">
+          <Alert.Heading className="text-danger">
+            <FormattedMessage
+              id="logistration.reset.password.request.server.error.header.message"
+              defaultMessage="Failed to Reset Password"
+              description="Failed to Reset Password help text heading."
+            />
+          </Alert.Heading>
+          <FormattedMessage
```
waheedahmed, line 17: I guess this will be a common error message for each page, we can move this into common components and pass the heading dynamically for each page.

Also, this message should appear above the form instead of the whole page.

waheedahmed (changes requested):

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -150,13 +154,15 @@ ResetPasswordPage.defaultProps = {
...
   errors: null,
+  errorCode: null,
 };
 
 ResetPasswordPage.propTypes = {
   intl: intlShape.isRequired,
   resetPassword: PropTypes.func.isRequired,
   validateToken: PropTypes.func.isRequired,
   token_status: PropTypes.string,
+  errorCode: PropTypes.string,
```
waheedahmed, line 165: Why we are always defining a new prop for each error case, can’t we utilize a single one for all?

src/reset-password/messages.js
```diff
@@ -61,6 +61,11 @@ const messages = defineMessages({
     defaultMessage: 'We couldn\'t reset your password.',
     description: 'Heading that appears above error message when user submits empty form.',
   },
+  'reset.password.request.server.error': {
+    id: 'reset.password.request.server.error',
+    defaultMessage: 'Failed to Reset Password',
```
waheedahmed, line 66: nit: `Failed to reset password`

src/reset-password/messages.js
```diff
@@ -61,6 +61,11 @@ const messages = defineMessages({
...
+    description: 'Failed to Reset Password help text heading.',
```
waheedahmed, line 67: nit: `Failed to reset password error message heading.`

waheedahmed (commented):
Don't use first letter capital cases within the sentences.

waheedahmed (changes requested):
I guess this should be handled for both token validation and password reset endpoints.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -78,22 +80,31 @@ const ResetPasswordPage = (props) => {
...
+  } else if (props.token_status === 'invalid' && props.errorCode !== INTERNAL_SERVER_ERROR) {
     return <InvalidTokenMessage />;
+  } else if (props.token_status === 'invalid' && props.errorCode === INTERNAL_SERVER_ERROR) {
+    return (
+      <div className="d-flex justify-content-center m-4">
+        <div className="d-flex flex-column mw-500">
+          <APIFailureMessage header={intl.formatMessage(messages['reset.password.token.validation.sever.error'])} />
+        </div>
+      </div>
+    );
```
waheedahmed, line 92: Move this condition after the below one so that you don't need to check for `props.errorCode === INTERNAL_SERVER_ERROR` in it.
```suggestion
  } else if (props.token_status === 'invalid' && props.errorCode === INTERNAL_SERVER_ERROR) {
    return (
      <div className="d-flex justify-content-center m-4">
        <div className="d-flex flex-column mw-500">
          <APIFailureMessage header={intl.formatMessage(messages['reset.password.token.validation.sever.error'])} />
        </div>
      </div>
    );
  } else if (props.token_status === 'invalid') {
    return <InvalidTokenMessage />;
```

src/reset-password/ResetPasswordPage.jsx
waheedahmed: I still encourage you to use existing status props instead of creating a new one for each case. I guess we can also refactor and combine the `token_status` and `status` as well into a single prop and use that for our purpose here as well.

PR #86, opened 2021-01-21
handle 500 on forgot password page

uzairr:
Screen Shot 2021-01-27 at 7 29 02 PM

VAN-281

### DIFF
--- a/src/forgot-password/data/reducers.js
+++ b/src/forgot-password/data/reducers.js
@@ -1,4 +1,5 @@
 import { FORGOT_PASSWORD } from './actions';
+import { INTERNAL_SERVER_ERROR } from '../../login/data/constants';
 
 export const defaultState = {
   status: null,
@@ -22,6 +23,11 @@ const reducer = (state = defaultState, action = null) => {
         return {
           ...state,
           status: 'forbidden',
+        };
+      case FORGOT_PASSWORD.FAILURE:
+        return {
+          ...state,
+          status: INTERNAL_SERVER_ERROR,
         };
       default:
         return state;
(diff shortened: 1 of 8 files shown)


src/forgot-password/data/sagas.js
```diff
@@ -20,9 +21,9 @@ export function* handleForgotPassword(action) {
     yield put(forgotPasswordSuccess(action.payload.email));
   } catch (e) {
     if (e.response && e.response.status === 403) {
-      yield put(forgotPasswordForbidden());
+      yield put(forgotPasswordForbidden(null));
     } else {
-      throw e;
+      yield put(forgotPasswordForbidden(INTERNAL_SERVER_ERROR));
```
waheedahmed, line 26: Why we are using forbidden action for 500 errors? It should have its own action and should utilities the `status` prop.

src/forgot-password/data/reducers.js
```diff
@@ -22,6 +22,7 @@ const reducer = (state = defaultState, action = null) => {
         return {
           ...state,
           status: 'forbidden',
+          errorCode: action.payload.errorCode,
```
waheedahmed, line 25: Instead of introducing a new property, we should utilities `status`.

waheedahmed (changes requested):
Same for this as well can create a common component for this type of error because it's going to be the same for all the pages?

src/forgot-password/APIFailureMessage.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { Alert } from '@edx/paragon';
+
+const APIFailureMessage = () => (
+  <div className="d-flex justify-content-center m-4">
+    <div className="d-flex flex-column">
+      <div className="text-center mw-500">
```
waheedahmed, line 8: Also, no need to wrap this message inside all these divs if it is going to appear at the top of the form.

adeelehsan (approved):
looks good.

PR #87, opened 2021-01-21
Adds frontend events for Google Analytics.

adeelkhan:
This patch adds tracking and page events destined for GA.
VAN-282

### DIFF
--- a/src/login/LoginHelpLinks.jsx
+++ b/src/login/LoginHelpLinks.jsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCaretDown, faCaretRight } from '@fortawesome/free-solid-svg-icons';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
@@ -22,8 +23,13 @@ const LoginHelpLinks = (props) => {
     setShowLoginHelpValue(!showLoginHelp);
   };
 
+  const handleForgotPasswordLinkClickEvent = () => {
+    sendTrackEvent('edx.bi.password-reset_form.toggled', { category: 'user-engagement' });
+  };
+
   const forgotPasswordLink = () => (
-    <a className="field-link" href={RESET_PAGE}>
+
+    <a className="field-link" href={RESET_PAGE} onClick={handleForgotPasswordLinkClickEvent}>
       {intl.formatMessage(messages['forgot.password.link'])}
     </a>
   );
--- a/src/login/LoginPage.jsx
+++ b/src/login/LoginPage.jsx
@@ -3,6 +3,7 @@ import Skeleton from 'react-loading-skeleton';
 
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { sendPageEvent, sendTrackEvent } from '@edx/frontend-platform/analytics';
 import {
   Form, Hyperlink, Input, StatefulButton, ValidationFormGroup,
 } from '@edx/paragon';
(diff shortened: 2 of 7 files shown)


src/logistration/LoginHelpLinks.jsx
```diff
@@ -22,8 +23,12 @@ const LoginHelpLinks = (props) => {
     setShowLoginHelpValue(!showLoginHelp);
   };
 
+  const handleForgotPasswordLinkClickEvent = () => {
+    sendGAEvent('analytics', 'edx.bi.password-reset_form.toggled', { category: 'user-engagement' });
+  };
+
   const forgotPasswordLink = () => (
-    <a className="field-link" href={RESET_PAGE}>
+    <a className="field-link" href={RESET_PAGE} onClick={(e) => handleForgotPasswordLinkClickEvent(e)}>
```
waheedahmed, line 31: I guess we don't need the click event `e` in `handleForgotPasswordLinkClickEvent` and remove it form the call?
```suggestion
    <a className="field-link" href={RESET_PAGE} onClick={handleForgotPasswordLinkClickEvent}>
```

src/logistration/RegistrationPage.jsx
```diff
@@ -242,6 +243,11 @@ class RegistrationPage extends React.Component {
     }
   }
 
+  handleLoginLinkClickEvent() {
+    sendGAEvent('analytics', 'edx.bi.login_form.toggled', { category: 'user-engagement' });
+    sendGAEvent('page', 'login_and_registration login');
```
waheedahmed, line 248: I guess `login` should go as a `type` param?

src/logistration/LoginPage.jsx
```diff
@@ -121,6 +124,11 @@ class LoginPage extends React.Component {
     });
   }
 
+  handleCreateAccountLinkClickEvent() {
+    sendGAEvent('analytics', 'edx.bi.register_form.toggled', { category: 'user-engagement' });
+    sendGAEvent('page', 'login_and_registration register');
```
waheedahmed, line 129: Same for the register, I'm not sure but I guess `register` should go as type in params? It's going as a separate param in old implementation https://github.com/edx/edx-platform/blob/5a09786cf3547a5d6ea395d4f7522c04efcb0a5f/lms/static/js/student_account/views/AccessView.js#L288.

src/logistration/LoginPage.jsx
```diff
@@ -56,6 +57,8 @@ class LoginPage extends React.Component {
   }
 
   handleInstitutionLogin = () => {
+    sendGAEvent('analytics', 'edx.bi.institution_login_form.toggled', { category: 'user-engagement' });
+    sendGAEvent('page', 'login_and_registration institution_login');
```
waheedahmed, line 61: And `institution_login` too.

src/logistration/LoginPage.jsx
adeelkhan: I think i was hardcoding the generated string ie pagename + type like **login_and_registration login**.  I am not sure how and for what purpose type is used in backbonejs and how they are combined to create above event name . My assumption is it is treated as category. https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#page
I have changed the code to reflect this now

adeelkhan:
@waheedahmed Please re-review.

src/logistration/LoginPage.jsx
```diff
@@ -57,6 +58,8 @@ class LoginPage extends React.Component {
   }
 
   handleInstitutionLogin = () => {
+    sendGAEvent('analytics', 'edx.bi.institution_login_form.toggled', null, { category: 'user-engagement' });
+    sendGAEvent('page', 'login_and_registration', 'institution_login');
```
waheedahmed, line 62: `login_and_registration` is the category and `login`, `register` and `institution_login` etc are the names of these events, I guess we are passing these here and on other places as the opposite.

waheedahmed (changes requested):
It seems like the page view events for the login and register pages no necessarily needed to be under the click event since our MFE has separate URLs and needs to be fired whenever the page is loaded. If we keep them inside the click event handler, they won't fire if the user opens the link directly by typing in the URL bar or coming from some other page.

The monolith uses this logic because it is built using Backbone and doesn't load the page every time the user clicks the link. I also verified that when the user lands directly on these pages in monolith FE, the event is get fired but with this logic it won't in our MFE.

adeelkhan:
@waheedahmed Here the changes based on above review. 
1) Removed call to dataUtils.js function.
2) Fixed page event call arguments based on new information.
3) Removed page events on click. Now as we have a form that gets refreshed rather than switched/togged, moving it inside render to cater for any possible flow via links, or page refresh.
4) Update tests.

waheedahmed (approved):

PR #90, opened 2021-01-23
Fix scrolling issue.

adeelkhan:
Scrolling was not working for error message
changes on form submission. useEffect was not
calling callback on its dependency change ie error message
.This patch would fix this.


waheedahmed (approved):
Can we also pass the behavior option to `window.scrollTo({ left: 0, top: 0, behavior: 'smooth' });`? Currently, it's suddenly scrolling to top.

PR #100, opened 2021-01-27
Adds UX suggested changes.

adeelkhan:
**Before:**
Screen Shot 2021-01-28 at 2 11 31 PM

**After:**
Screen Shot 2021-01-29 at 3 58 26 PM

VAN-324


waheedahmed (commented):
I guess the social and university SSO buttons meant to be under the `Create Account` button? Otherwise, changes looks good to me.

adeelehsan:
@adeelkhan 1px border after `already have an account` is also missing. Please add that as well

waheedahmed (approved):

PR #106, opened 2021-01-29
Zamir/van 299/fix reset password page bugs

zainab-amir:
### Notes:
- Added empty form error message 
- Fixed stateful button
- Take page focus to top of page when rate limit error is shown
- Use same color header as the body

Screenshot 2021-01-29 at 5 43 52 PM

[VAN-299](https://openedx.atlassian.net/browse/VAN-299)

### DIFF
--- a/src/reset-password/messages.js
+++ b/src/reset-password/messages.js
@@ -36,6 +36,31 @@ const messages = defineMessages({
     defaultMessage: 'Password Reset Complete.',
     description: 'header message when reset is successful.',
   },
+  'forgot.password.confirmation.sign.in.link': {
+    id: 'forgot.password.confirmation.sign.in.link',
+    defaultMessage: 'sign-in',
+    description: 'link text used in message to refer to sign in page',
+  },
+  'reset.password.request.forgot.password.text': {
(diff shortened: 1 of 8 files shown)


src/reset-password/messages.js
```diff
@@ -36,6 +36,31 @@ const messages = defineMessages({
...
+    description: 'Forgot password text',
+  },
+  'reset.password.request.invalid.token.header': {
+    id: 'reset.password.request.invalid.token.header',
+    defaultMessage: 'Invalid Password Reset Link',
+    description: 'Invalid password reset link help text heading',
+  },
+  'reset.password.empty.new.password.field.error': {
+    id: 'reset.password.empty.new.password.field.error',
+    defaultMessage: 'Please enter your New Password.',
```
waheedahmed, line 56: Are we intentionally keeping the first letters capital for `New Password`?

waheedahmed (approved):
Just one question otherwise 👍

src/reset-password/messages.js
zainab-amir: yes to match the field name

PR #111, opened 2021-02-01
Remove margin bottom from register fields

adeelkhan:
Screen Shot 2021-02-01 at 1 27 08 PM

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -403,7 +403,7 @@ class RegistrationPage extends React.Component {
                 <ValidationFormGroup
                   for={field.name}
                   key={field.name}
-                  className="pt-10 mb-20 small"
+                  className="pt-10 small"
                 >
                   { nodes }
                 </ValidationFormGroup>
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -403,7 +403,7 @@ class RegistrationPage extends React.Component {
                 <ValidationFormGroup
                   for={field.name}
                   key={field.name}
-                  className="pt-10 mb-20 small"
+                  className="pt-10 mb-0 small"
```
waheedahmed, line 406: No need to add `mb-0` class, we need to keep `1rem` default `margin-bottom` set on `.form-group`.

waheedahmed (commented):
Remove `mb-0`, we need the default `margin-bottom` set on `.form-group`

src/register/RegistrationPage.jsx
```diff
@@ -683,7 +679,7 @@ class RegistrationPage extends React.Component {
                 { this.addExtraRequiredFields() }
                 <ValidationFormGroup
                   for="optional"
-                  className="custom-control pt-10 mb-0"
```
waheedahmed, line 686: This shouldn’t be removed. We need zero bottom margin on this.

adeelkhan: done

src/register/RegistrationPage.jsx
adeelkhan: done

waheedahmed (approved):

PR #117, opened 2021-02-02
password page email validation removed

adeelehsan:
van-316
Email validation removed `onChange` event.
Paragon helper text added.
Updated the error message incase of invalid email
Screen Shot 2021-02-02 at 12 15 15 PM

### DIFF
--- a/src/forgot-password/ForgotPasswordPage.jsx
+++ b/src/forgot-password/ForgotPasswordPage.jsx
@@ -8,8 +8,10 @@ import {
   Input,
   StatefulButton,
   ValidationFormGroup,
+  Alert,
 } from '@edx/paragon';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { getConfig } from '@edx/frontend-platform';
 
 import { Formik } from 'formik';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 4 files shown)


src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -25,22 +26,34 @@ import { INTERNAL_SERVER_ERROR } from '../login/data/constants';
 
 const ForgotPasswordPage = (props) => {
   const { intl, status } = props;
-  let invalidEmailMessage;
+  const [isEmailValid, setIsEmailValid] = useState(true);
```
waheedahmed, line 29: Instead of the state, can we keep using formik?

src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -50,27 +66,34 @@ const ForgotPasswordPage = (props) => {
...
     <Formik
-      onSubmit={(values) => {
-        if (values.isEmailValid) {
-          props.forgotPassword(values.email);
+      validate={(values) => {
+        const regex = new RegExp(VALID_EMAIL_REGEX, 'i');
+        if (!regex.test(values.email)) {
+          return { email: intl.formatMessage(messages['forgot.password.page.invalid.email.message']) };
         }
+        props.forgotPassword(values.email);
```
waheedahmed, line 74: Why we are calling the action in `validate` function? Can we keep this in `onSubmit`?

src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -25,22 +31,32 @@ import { INTERNAL_SERVER_ERROR } from '../login/data/constants';
...
-  let invalidEmailMessage;
 
-  const validateEmail = (e, setFieldValue) => {
-    invalidEmailMessage = intl.formatMessage(messages['forgot.password.page.invalid.email.message']);
-    const regex = new RegExp(VALID_EMAIL_REGEX, 'i');
+  const platformName = getConfig().SITE_NAME;
+  const emailHelpMessage = (
+    <FormattedMessage
+      id="forgot.password.email.help.text"
+      defaultMessage="The email address you used to register with {platformName}."
```
waheedahmed, line 39: Move this into `messages.jsx` and then you can assign `helpText=intl.formatMessage(messages['forgot.password.email.help.text'])` and no need to create a variable.

waheedahmed: You can pass the values to `intl.formatMessage()` as well like we are doing here. https://github.com/edx/frontend-app-authn/blob/667389b1312375e3a08c546e8de138882b127342/src/login/LoginFailure.jsx#L43

waheedahmed (approved):

PR #131, opened 2021-02-09
Fix buttons needs to click twice issue.

waheedahmed:
The `onBlur` event's priority is higher than `onClick` and rerenders the component before the click event fires.

VAN-352

## Testing:
Sandbox: https://authn-van.sandbox.edx.org/login


zainab-amir (approved):

zainab-amir (approved):
Does this work if we submit empty form without any previous onBlur event?

waheedahmed:
@zainab-amir Yes it does work with an empty form, the sandbox is up to date; you can test it out.

PR #140, opened 2021-02-15
Add multiple enterprise redirect

zainab-amir:
If redirect url has auth complete url in it, it means that the redirect url is for multiple enterprise selection page and user must be redirected to that page first.

example url = http://localhost:18000/enterprise/select/active/?success_url=/auth/complete/google-oauth2/

This check mimics the [following scenario](https://github.com/edx/edx-platform/blob/50bb70298cd30170e5572dfd9e7d2914dec3a4cd/lms/static/js/student_account/views/AccessView.js#L319-L320) from FE:
```javascript
 loginComplete: function() {
    if (this.thirdPartyAuth && this.thirdPartyAuth.finishAuthUrl) {
        multipleEnterpriseInterface.check(this.thirdPartyAuth.finishAuthUrl);
        // Note: the third party auth URL likely contains another redirect URL embedded inside
    } else {
        multipleEnterpriseInterface.check(this.nextUrl);
    }
 },
```

[VAN-311](https://openedx.atlassian.net/browse/VAN-311)

### DIFF
--- a/src/common-components/RedirectLogistration.jsx
+++ b/src/common-components/RedirectLogistration.jsx
@@ -4,12 +4,14 @@ import PropTypes from 'prop-types';
 import { getConfig } from '@edx/frontend-platform';
 
 function RedirectLogistration(props) {
-  const { finishAuthUrl, success, redirectUrl } = props;
+  const { finishAuthUrl, redirectUrl, success } = props;
 
   if (success) {
     // If we're in a third party auth pipeline, we must complete the pipeline
     // once user has successfully logged in. Otherwise, redirect to the specified redirect url.
-    if (finishAuthUrl) {
+    // Note: For multiple enterprise use case, we need to make sure that user first visits the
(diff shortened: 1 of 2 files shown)


src/common-components/RedirectLogistration.jsx
```diff
@@ -10,7 +10,13 @@ function RedirectLogistration(props) {
     // If we're in a third party auth pipeline, we must complete the pipeline
     // once user has successfully logged in. Otherwise, redirect to the specified redirect url.
     if (finishAuthUrl) {
```
waheedahmed, line 12: I guess we could combine the second condition with a not (!) operator here:
```suggestion
    if (finishAuthUrl && !redirectUrl.includes(finishAuthUrl)) {
```

waheedahmed (commented):
I guess we could combine the if condition here instead of another if/else inside it.

waheedahmed (approved):

PR #141, opened 2021-02-15
Performance optimisation

waheedahmed:
Also refactored page event logic.

VAN-343

**Sandbox:** https://authn-van.sandbox.edx.org/login

**Performance stats after the changes:** 
https://speedcurve.com/edx/favorite/?cs=lg&d=3&db=272893&dc=2&de=1&df=%7B%7D&ds=1


src/forgot-password/LoadableForgotPasswordPage.jsx
```diff
@@ -0,0 +1,11 @@
...
+import loadable from '@loadable/component';
+import Spinner from '../common-components/Spinner';
+
+const LoadableComponent = loadable(() => import('./ForgotPasswordPage.jsx'), {
+  fallback: <Spinner />,
+});
+
+const LoadableForgotPasswordPage = () => <LoadableComponent />;
+
+export default LoadableForgotPasswordPage;
```
zainab-amir, line 11: do we need separate component for it? Can't we add this code directly below the ForgotPasswordPage.jsx and just export that instead of exporting `ForgotPasswordPage`

src/forgot-password/LoadableForgotPasswordPage.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import loadable from '@loadable/component';
+import Spinner from '../common-components/Spinner';
+
+const LoadableComponent = loadable(() => import('./ForgotPasswordPage.jsx'), {
+  fallback: <Spinner />,
```
adeelehsan, line 6: I think instead of spinner we can use bar like github

adeelehsan (changes requested):

src/common-components/UnAuthOnlyRoute.jsx
```diff
@@ -10,12 +12,24 @@ import { DEFAULT_REDIRECT_URL } from '../data/constants';
  */
 const UnAuthOnlyRoute = (props) => {
```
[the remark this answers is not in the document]
waheedahmed: @adeelehsan this is used for unauthenticated routes only that's why it's named like this.

src/forgot-password/LoadableForgotPasswordPage.jsx
waheedahmed: Since we are using the spinner in other places as well like during token validation etc, I guess it doesn’t make sense to implement a new loader here. That's a good suggestion ad we can consider changing it later on after the launch.

src/forgot-password/index.js
```diff
@@ -1,4 +1,4 @@
-export { default } from './ForgotPasswordPage';
+export { default } from './LoadableForgotPasswordPage';
```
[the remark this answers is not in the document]
waheedahmed: I guess not, both are different components.

src/common-components/UnAuthOnlyRoute.jsx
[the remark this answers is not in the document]
waheedahmed: @adeelehsan renamed to `UnAuthenticatedRoute`.

src/common-components/UnAuthOnlyRoute.jsx
```diff
@@ -10,12 +12,24 @@ import { DEFAULT_REDIRECT_URL } from '../data/constants';
...
   const { authenticatedUser, config } = React.useContext(AppContext);
+  const match = useRouteMatch({
+    path: props.path,
+    exact: props.exact,
+    strict: props.strict,
+    sensitive: props.sensitive,
+  });
```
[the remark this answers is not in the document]
waheedahmed: I guess there is no readability issue, also don't want to create an extra variable.

src/forgot-password/LoadableForgotPasswordPage.jsx
waheedahmed: @zainab-amir I guess not, because in that case, the app will load the actual code as well while importing all these components in `index.jsx` which actually we want to avoid here.

waheedahmed: It also makes code cleaner to keep them separate.

zainab-amir (approved):

PR #143, opened 2021-02-16
Removes option call from validation endpoint

adeelkhan:
This would removes unnecessary csrf token and options calls from multiple endpoints.

VAN-370


src/register/data/service.js
```diff
@@ -6,6 +6,7 @@ export async function registerRequest(registrationInformation) {
   const requestConfig = {
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
     isPublic: true,
+    isCsrfExempt: true,
```
adeelehsan, line 9: please change it to httpClient

adeelkhan: Done!.

adeelkhan:
@adeelehsan Please re-review

adeelehsan (approved):

PR #146, opened 2021-02-17
backend validations updated for registration form

adeelehsan:
VAN-376

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -50,15 +50,13 @@ class RegistrationPage extends React.Component {
       goals: '',
       levelOfEducation: '',
       enableOptionalField: false,
-      validationErrorsAlertMessages: {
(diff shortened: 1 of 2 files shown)


waheedahmed (approved):

src/register/RegistrationPage.jsx
```diff
@@ -216,18 +208,31 @@ class RegistrationPage extends React.Component {
...
-      assignRegistrationErrorsToField: false,
+      updateFieldErrors: false,
+      updateAlertErrors: false,
+    }, () => {
+      this.validateInput(name, value, payload, false);
     });
+  }
 
-    this.props.fetchRealtimeValidations(payload);
+  handleOnChange(e) {
```
zainab-amir, line 220: handleOnChange is not used any where is code. Need to update the onChange handler for `AuthnValidationFormGroup`

PR #147, opened 2021-02-17
make registration form static

zainab-amir:
### WIP

make registration form static

- removed fetch registration call from register page
- removed unused state variables
- moved few hardcoded links to env variables

[VAN-369](https://openedx.atlassian.net/browse/VAN-369)


src/register/RegistrationPage.jsx
```diff
@@ -744,7 +605,41 @@ class RegistrationPage extends React.Component {
...
+                  invalid={this.state.errors.country !== ''}
+                  invalidMessage={intl.formatMessage(messages['country.validation.message'])}
+                  className="mb-0"
+                  value={this.state.country}
+                  onClick={(e) => this.handleOnClick(e)}
+                  onBlur={(e) => this.handleOnBlur(e)}
+                  onChange={(e) => this.handleOnChange(e)}
+                  selectOptions={this.getOptions().countryOptions}
+                />
+                <input id="honorCode" name="honorCode" type="hidden" value={this.state.honorCode} required onChange={e => this.handleOnChange(e)} />
```
adeelkhan, line 623: i think backend need id to "honor_code"

src/register/RegistrationPage.jsx
```diff
@@ -83,16 +76,7 @@ class RegistrationPage extends React.Component {
         username: '',
         password: '',
         country: '',
-        honorCode: '',
```
adeelkhan, line 86: Removing this would was cause failure in `handleOnChange` for honorCode. 
`[e.target.name]: targetValue,`

src/register/RegistrationPage.jsx
adeelkhan, line 623: honor code should always be true and i don't think we are allowing user to change its state. The honor code paragraph is meant to be an implicit agreement by user

adeelkhan:
Rest looks good.

src/register/RegistrationPage.jsx
zainab-amir: I will remove this and keep it as simple text. Removing this will also fix your concern mentioned above

src/register/RegistrationPage.jsx
```diff
@@ -147,31 +116,98 @@ class RegistrationPage extends React.Component {
...
+        label: intl.formatMessage(messages[`registration.field.education.levels.${key || 'label'}`]),
+      })),
+      genderOptions: GENDER_OPTIONS.map(key => ({
+        value: key,
+        label: intl.formatMessage(messages[`registration.field.gender.options.${key || 'label'}`]),
+      })),
+    };
+  };
+
+  getOptionalFields() {
```
waheedahmed, line 141: Can we move this into a separate component `OptionalFields` instead of a function to make this code cleaner and shorter?

waheedahmed (approved):
One question otherwise 👍

src/register/RegistrationPage.jsx
zainab-amir: how will we get it's value to send in form payload? We will react to use redux for it. wouldn't it make it complicated? Let me know if I am thinking in the wrong direction

zainab-amir: Now that I think about it, I think it can be achieved by just sending onChange event to the component. Let me refactor the code

waheedahmed (approved):
Looks great, nice work 👍

PR #149, opened 2021-02-19
fix optional field not available

zainab-amir:
if `REGISTRATION_OPTIONAL_FIELDS` is empty, do not render optional field button

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -528,20 +528,22 @@ class RegistrationPage extends React.Component {
                     }}
                   />
                 </div>
-                <AuthnValidationFormGroup
-                  label=""
-                  for="optional"
-                  name="optional"
-                  type="checkbox"
-                  value={this.state.enableOptionalField}
-                  onClick={(e) => this.handleOnOptional(e)}
-                  onBlur={null}
-                  onChange={(e) => this.handleOnOptional(e)}
-                  optionalFieldCheckbox
-                  isChecked={this.state.enableOptionalField}
-                  checkboxMessage={intl.formatMessage(messages['support.education.research'])}
-                />
-                { this.state.enableOptionalField ? this.getOptionalFields() : null}
+                {getConfig().REGISTRATION_OPTIONAL_FIELDS ? (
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -528,20 +528,22 @@ class RegistrationPage extends React.Component {
...
+                    value={this.state.enableOptionalField}
+                    onClick={(e) => this.handleOnOptional(e)}
+                    onBlur={null}
+                    onChange={(e) => this.handleOnOptional(e)}
+                    optionalFieldCheckbox
+                    isChecked={this.state.enableOptionalField}
+                    checkboxMessage={intl.formatMessage(messages['support.education.research'])}
+                  />
+                ) : null}
+                { this.state.enableOptionalField ? this.getOptionalFields() : null }
```
waheedahmed, line 546: Shouldn't we move this as well under this check?

waheedahmed (approved):
One question otherwise 👍

src/register/RegistrationPage.jsx
zainab-amir: the default value for `enableOptionalField` is false so it will never be rendered

src/register/tests/RegistrationPage.test.jsx
```diff
@@ -145,6 +145,18 @@ describe('RegistrationPageTests', () => {
...
+  it('should not show optional field check if process env has empty optional fields list', () => {
+    mergeConfig({
+      REGISTRATION_OPTIONAL_FIELDS: '',
+    });
+    const registrationPage = mount(reduxWrapper(<IntlRegistrationPage {...props} />));
+    expect(registrationPage.find('input#optional').length).toEqual(0);
+
+    mergeConfig({
+      REGISTRATION_OPTIONAL_FIELDS: 'gender goals level_of_education year_of_birth',
+    });
```
waheedahmed, line 157: Remove it or add an assert after  it for success case:
```
expect(registrationPage.find('input#optional').length).toEqual(1);
```

waheedahmed:
:shipit:

PR #150, opened 2021-02-19
VAN-393: use sentence case for text

zainab-amir:
Register page:
- [x] Update placeholders on the register page to use sentence case.
- [x] Update button test to use sentence case. 

Login page:
- [x] Update title to use sentence case.
- [x] update "Create an Account" hyperlink to use sentence case. 

Password Assistance
- [x] update hyperlink "Other sign-in issues" and replace hyphen with space.
- [x]  update "Check your Email" title to use sentence case for when user has requested a password recovery. 

Reset Password page:
- [x] Update placeholders on reset password page to use sentence case.

[VAN-393](https://openedx.atlassian.net/browse/VAN-393)

### DIFF
--- a/src/register/messages.jsx
+++ b/src/register/messages.jsx
@@ -3,7 +3,7 @@ import { defineMessages } from '@edx/frontend-platform/i18n';
 const messages = defineMessages({
   'create.account.button': {
     id: 'create.account.button',
-    defaultMessage: 'Create Account',
+    defaultMessage: 'Create account',
     description: 'Button label that appears on register page',
   },
   'already.have.an.edx.account': {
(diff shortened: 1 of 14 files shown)


src/register/messages.jsx
```diff
@@ -123,7 +123,7 @@ const messages = defineMessages({
   },
   'support.education.research': {
     id: 'support.education.research',
-    defaultMessage: 'Support education research by providing additional information. (Optional)',
+    defaultMessage: 'Support education research by providing additional information. (optional)',
```
waheedahmed, line 126: Shouldn’t we keep it capital here? It's after the dot.

waheedahmed (approved):
Just one question otherwise 👍

src/register/messages.jsx
zainab-amir: good catch. will confirm from @wajeeha-khalid  to be sure

PR #169, opened 2021-02-25
Revert "Performance optimisation"

waheedahmed:
Reverts edx/frontend-app-authn#141


waheedahmed:
It's breaking some flows, reverting.

PR #174, opened 2021-03-01
react helmet package added to change each page title

adeelehsan:
VAN-335

### DIFF
--- a/src/forgot-password/ForgotPasswordPage.jsx
+++ b/src/forgot-password/ForgotPasswordPage.jsx
@@ -3,6 +3,7 @@ import React, { useState } from 'react';
 import { Formik } from 'formik';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { Helmet } from 'react-helmet';
 import { Redirect } from 'react-router-dom';
 
 import { getConfig } from '@edx/frontend-platform';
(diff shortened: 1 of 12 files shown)


src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -86,6 +87,10 @@ const ForgotPasswordPage = (props) => {
         errors, handleSubmit, setFieldValue, values,
       }) => (
         <>
+          <Helmet>
+            <meta charSet="utf-8" />
```
waheedahmed, line 91: Is it necessary, I guess we can remove this?

src/login/LoginPage.jsx
```diff
@@ -174,6 +175,10 @@ class LoginPage extends React.Component {
 
     return (
       <>
+        <Helmet>
+          <meta charSet="utf-8" />
```
waheedahmed, line 179: Same here, remove this if not necessary.

waheedahmed (commented):
Also, mark all title strings for translation.

src/reset-password/ResetPasswordPage.jsx
```diff
@@ -120,7 +121,9 @@ const ResetPasswordPage = (props) => {
   } else {
     return (
       <>
-
+        <Helmet>
```
waheedahmed, line 124: I guess this needs to be added on invalid token and password reset success cases as well.

waheedahmed (approved):

PR #179, opened 2021-03-03
update registration failure message

zainab-amir:
Changes made in https://github.com/edx/edx-platform/pull/26821 will update how the backend returns response. This PR makes sure that backend changes don't break the registration workflow on MFE.

### DIFF
--- a/src/register/RegistrationFailure.jsx
+++ b/src/register/RegistrationFailure.jsx
@@ -28,8 +28,9 @@ const RegistrationFailureMessage = (props) => {
      userErrors.push(serverError);
      break;
 
-     default:
-        Object.keys(errorMessage).forEach((key) => {
+    default:
+      Object.keys(errorMessage).forEach((key) => {
+        if (key !== 'error_code') {
           const errors = errorMessage[key];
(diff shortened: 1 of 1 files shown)


src/register/RegistrationFailure.jsx
```diff
@@ -28,8 +28,9 @@ const RegistrationFailureMessage = (props) => {
      userErrors.push(serverError);
      break;
 
-     default:
-        Object.keys(errorMessage).forEach((key) => {
+    default:
+      Object.keys(errorMessage).forEach((key) => {
+        if (key !== 'error_code') {
```
adeelkhan, line 33: Do we have backend cases where the key fields has error_code?  Does this inject error_code into fields?
https://github.com/edx/edx-platform/pull/26821/files?file-filters%5B%5D=.py#diff-edb3995ebb3452c122e32954149f176d38476d0c6b960164d09ff48f8d9a5f51R657

zainab-amir: In the response dictionary it saves a new key (error_code) with it's value set to the type of error.

adeelkhan (approved):

PR #205, opened 2021-03-16
Track total time it takes users to register

zainab-amir:
Added a new param to payload in registration page. This tracks the total time it took user to register.

VAN-413

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -69,6 +69,7 @@ class RegistrationPage extends React.Component {
       },
       institutionLogin: false,
       formValid: false,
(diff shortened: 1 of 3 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -188,6 +190,7 @@ class RegistrationPage extends React.Component {
       }
     });
     if (finalValidation) {
+      payload.totalRegistrationTime = totalRegistrationTime.toString();
```
waheedahmed, line 193: Why do we need to convert it into a string? Also, can we also do the calculations here once all validations pass?

waheedahmed (approved):
👍 except for one comment.

PR #212, opened 2021-03-18
Handle ratelimit error for registration endpoint

zainab-amir:
As part of adding ratelimit to registration, we need to make sure MFE works and shows error messages as expected.

Screenshot 2021-03-18 at 3 25 38 PM

Note: Currently I am ignoring the openedX `403` error regarding [required third party auth](https://github.com/edx/edx-platform/blob/3cdbb5b36aaf81b0459823db9a2c9ca40f9d7b7a/openedx/core/djangoapps/user_authn/views/register.py#L513-L517) and displaying a generic message _"too many failed attempts"_. Once we add support for openedX we need to make sure the appropriate error message is show for different `403` errors.

Tickert: [VAN-302](https://openedx.atlassian.net/browse/VAN-302)

### DIFF
--- a/src/register/RegistrationFailure.jsx
+++ b/src/register/RegistrationFailure.jsx
@@ -1,10 +1,12 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
+
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Alert } from '@edx/paragon';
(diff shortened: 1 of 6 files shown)


waheedahmed (commented):
Please see my comment on platform PR and update it accordingly.

waheedahmed (approved):

PR #222, opened 2021-03-26
Revert "Add 3:1 contrast to all form text field."

waheedahmed:
Reverts edx/frontend-app-authn#220

Latest paragon version causing some issues with the current implementation.


mubbsharanwar (approved):
revert @edx/paragon version

PR #228, opened 2021-03-30
skip hinted login check added for enterprise SSO

adeelehsan:


src/data/utils/dataUtils.js
```diff
@@ -24,7 +24,9 @@ export const getTpaProvider = (tpaHintProvider, primaryProviders, secondaryProvi
     secondaryProviders.forEach((provider) => {
       if (provider.id === tpaHintProvider) {
         tpaProvider = provider;
-        isSecondaryProvider = true;
+        if (provider.skipHintedLogin) {
```
waheedahmed, line 27: This needs in above for each as well, these providers can be primary as well.

I also suggest concatenating both arrays into one and change this function to use only one for each loop instead of two because there is no difference in the logic for both.

src/data/utils/dataUtils.js
```diff
@@ -24,7 +24,9 @@ export const getTpaProvider = (tpaHintProvider, primaryProviders, secondaryProvi
...
+          isSecondaryProvider = true;
```
waheedahmed, line 28: Since this is not necessarily a secondary provider we should change the variable name to `skipHintedLogin` or `redirectToIdp` etc.

waheedahmed (changes requested):

waheedahmed (approved):

PR #233, opened 2021-04-06
Send user attributes to hotjar

zainab-amir:
Once the user has signed up successfully, send custom user attribute to hotjar. These can be used to configure hotjar surveys and other feedback tools. 

#### Other information:
- I am following this guide to setup user attributes: https://help.hotjar.com/hc/en-us/articles/360038394053/
- I am not sending edX user id to identify call and it will automatically use the hotjar user id. 
- **Test Account** attribute will be removed after testing 

[VAN-291](https://openedx.atlassian.net/browse/VAN-291)

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -432,6 +432,13 @@ class RegistrationPage extends React.Component {
       );
     }
 
+    if (this.props.registrationResult.success && window.hj) {
+      window.hj('identify', null, {
+        signedUp: new Date().toISOString(),
+        testAccount: this.state.username.includes('hotjarTest'), // TODO: Remove this after testing
+      });
+    }
+
     return (
       <>
         <Helmet>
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -432,6 +432,13 @@ class RegistrationPage extends React.Component {
       );
     }
 
+    if (this.props.registrationResult.success && window.hj) {
+      window.hj('identify', null, {
+        'Signed up': new Date().toISOString(),
+        'Test Account': this.state.username.includes('hotjarTest'), // TODO: Remove this after testing
```
waheedahmed, line 438: Can we always send the username attribute and add that condition on HJ side?

src/register/RegistrationPage.jsx
```diff
@@ -432,6 +432,13 @@ class RegistrationPage extends React.Component {
       );
     }
 
+    if (this.props.registrationResult.success && window.hj) {
+      window.hj('identify', null, {
+        'Signed up': new Date().toISOString(),
```
waheedahmed, line 437: Use camel case for attribute names.

src/register/RegistrationPage.jsx
zainab-amir: I am not sure if we should send any PII, even in test case I am sending whether it is test user or not (true or false)

waheedahmed: Hmm, makes sense 👍

waheedahmed (approved):

PR #234, opened 2021-04-06
Make optional fields configurable

zainab-amir:
As part of VAN-405, changes are required to remove the optional fields, instead of removing them completely from code base I have made them configurable which will allow us to enable it back if we need it. If it's not needed in future we can remove it completely.

##### Empty `REGISTRATION_OPTIONAL_FIELDS` process env gives us:
Screenshot 2021-04-06 at 5 57 09 PM

I have removed custom class from the fields because enabling some fields was causing css to break
Screenshot 2021-04-06 at 5 48 45 PM
 

[VAN-405](https://openedx.atlassian.net/browse/VAN-405)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -23,4 +23,4 @@ AUTHN_MINIMAL_HEADER=true
 LOGIN_ISSUE_SUPPORT_LINK=''
 TOS_AND_HONOR_CODE='http://localhost:18000/honor'
 PRIVACY_POLICY='http://localhost:18000/privacy'
-REGISTRATION_OPTIONAL_FIELDS='gender,goals,level_of_education,year_of_birth'
+REGISTRATION_OPTIONAL_FIELDS='gender,goals,levelOfEducation,yearOfBirth'
--- a/src/register/OptionalFields.jsx
+++ b/src/register/OptionalFields.jsx
@@ -9,7 +9,9 @@ import messages from './messages';
 import { AuthnValidationFormGroup } from '../common-components';
 
 const OptionalFields = (props) => {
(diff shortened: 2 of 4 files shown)


.env.development
```diff
@@ -23,4 +23,4 @@ AUTHN_MINIMAL_HEADER=true
 LOGIN_ISSUE_SUPPORT_LINK=''
 TOS_AND_HONOR_CODE='http://localhost:18000/honor'
 PRIVACY_POLICY='http://localhost:18000/privacy'
-REGISTRATION_OPTIONAL_FIELDS='gender,goals,level_of_education,year_of_birth'
+REGISTRATION_OPTIONAL_FIELDS='{"gender":true,"goals":true,"levelOfEducation":true,"yearOfBirth":true}'
```
waheedahmed, line 26: This works locally but not sure if it will with `frontend_build` script in tubular. We had a problem with this variable in the past with space-separated values.

waheedahmed: Instead of using a true/false, can’t we simply remove the values from `REGISTRATION_OPTIONAL_FIELDS` array to hide them from the form? If there is no field specified in it, you could simply hide all the content related to optional fields.

waheedahmed (changes requested):

.env.development
zainab-amir: yes it can work, I will update it

waheedahmed (approved):

zainab-amir:
Note: Going to merge this PR even though "codecov/project" is failing right now. I am going to be creating a follow up PR for VAN-405 where I will be changing existing tests/writing new test for register page

PR #237, opened 2021-04-09
Update optional fields to use updated paragon components

zainab-amir:
Added back compatibility for optional fields by updating them to new paragon components, these can be used if we decide to enable optional fields in future.
Screenshot 2021-04-09 at 5 43 42 PM

#### Changes
- updated paragon package
- updated optional fields to use new paragon components
- some tests needed fixing after updating to new paragon component

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@edx/frontend-component-cookie-policy-banner": "2.1.11",
     "@edx/frontend-component-header": "2.2.4",
     "@edx/frontend-platform": "1.8.4",
-    "@edx/paragon": "13.16.1",
+    "@edx/paragon": "13.17.5",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "@edx/frontend-component-cookie-policy-banner": "2.1.11",
     "@edx/frontend-component-header": "2.2.4",
     "@edx/frontend-platform": "1.8.4",
-    "@edx/paragon": "13.16.1",
+    "@edx/paragon": "^13.17.5",
```
waheedahmed, line 41: Remove the caret sign.

waheedahmed (approved):
One minor comment otherwise 👍

package.json
zainab-amir: okay

PR #242, opened 2021-04-13
Base component unit test

mubbsharanwar:
Write this basic test to complete test coverage 100%.

### DIFF
--- a/src/common-components/tests/BaseComponent.test.jsx
+++ b/src/common-components/tests/BaseComponent.test.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { IntlProvider } from '@edx/frontend-platform/i18n';
+
+import SmallScreenLayout from '../SmallScreenLayout';
+import MediumScreenLayout from '../MediumScreenLayout';
+import LargeScreenLayout from '../LargeScreenLayout';
+
+describe('ScreenLayout', () => {
+  it('should display the form, pass as a child in SmallScreenLayout', () => {
+    const smallScreen = mount(
+      <IntlProvider locale="en">
+        <SmallScreenLayout>
+          <form>
+            <input type="text" />
+          </form>
+        </SmallScreenLayout>
+      </IntlProvider>,
+    );
+    expect(smallScreen.find('form').exists()).toEqual(true);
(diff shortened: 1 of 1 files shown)


src/common-components/tests/BaseComponent.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const smallScreen = mount(
+      <IntlProvider locale="en">
+        <SmallScreenLayout>
+          <form>
+            <input type="text" />
+          </form>
+        </SmallScreenLayout>
+      </IntlProvider>,
+    );
+    expect(smallScreen.find('form').exists());
```
waheedahmed, line 20: This statement itself does not verify anything, if `exists()` returned false this test will not fail. You need to verify it against true like this:
```suggestion
    expect(smallScreen.find('form').exists()).toEqual(True);
```

src/common-components/tests/BaseComponent.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const mediumScreen = mount(
+      <IntlProvider locale="en">
+        <MediumScreenLayout>
+          <form>
+            <input type="text" />
+          </form>
+        </MediumScreenLayout>
+      </IntlProvider>,
+    );
+    expect(mediumScreen.find('form').exists());
```
waheedahmed, line 33: Same here 
```suggestion
    expect(mediumScreen.find('form').exists()).toEqual(True);
```

src/common-components/tests/BaseComponent.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const largeScreen = mount(
+      <IntlProvider locale="en">
+        <LargeScreenLayout>
+          <form>
+            <input type="text" />
+          </form>
+        </LargeScreenLayout>
+      </IntlProvider>,
+    );
+    expect(largeScreen.find('form').exists());
```
waheedahmed, line 46: And here
```suggestion
    expect(largeScreen.find('form').exists()).toEqual(True);
```

waheedahmed (changes requested):

waheedahmed (approved):
Squash your commits before merging.

PR #247, opened 2021-04-14
Cookie added for cta dialogue

adeelehsan:


src/common-components/RedirectLogistration.jsx
```diff
@@ -11,6 +12,9 @@ function RedirectLogistration(props) {
     // once user has successfully logged in. Otherwise, redirect to the specified redirect url.
     // Note: For multiple enterprise use case, we need to make sure that user first visits the
     // enterprise selection page and then complete the auth workflow
+
+    const cookies = new Cookies();
+    cookies.set('edx-show-cta-dialogue', true, { path: '/' });
```
waheedahmed, line 17: This component is used in login case as well, I don't think this is the right place for this. Also, move the cookie name into some environment variable.

waheedahmed (changes requested):
The CTA itself does not represent what it is used for, might need to rephrase it to something `edx-show-activate-cta`.

waheedahmed:
Or maybe we don't need to set it here and instead in the registration API endpoint in LMS upon successful registration.

PR #251, opened 2021-04-20
Pre populate country

mubbsharanwar:
- get country code based on IP address from server
- set country automatically based on IP address in registration form

VAN-366

### DIFF
--- a/src/common-components/data/service.js
+++ b/src/common-components/data/service.js
@@ -11,7 +11,7 @@ export async function getThirdPartyAuthContext(urlParams) {
 
   const { data } = await getAuthenticatedHttpClient()
     .get(
-      `${getConfig().LMS_BASE_URL}/api/third_party_auth_context`,
+      `${getConfig().LMS_BASE_URL}/api/mfe_context`,
       requestConfig,
     )
     .catch((e) => {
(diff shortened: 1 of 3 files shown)


src/register/tests/RegistrationPage.test.jsx
```diff
@@ -627,6 +628,37 @@ describe('RegistrationPage', () => {
       mount(reduxWrapper(<IntlRegistrationPage {...props} />));
       expect(analytics.sendPageEvent).toHaveBeenCalledWith('login_and_registration', 'register');
     });
+
+    it('should select country based on IP address', () => {
```
waheedahmed, line 632: nit: rename the test to `Should update state if countryCode is present in context.`

src/common-components/data/service.js
```diff
@@ -11,7 +11,7 @@ export async function getThirdPartyAuthContext(urlParams) {
 
   const { data } = await getAuthenticatedHttpClient()
     .get(
-      `${getConfig().LMS_BASE_URL}/api/third_party_auth_context`,
+      `${getConfig().LMS_BASE_URL}/api/mfe_app_auth_context`,
```
waheedahmed, line 14: nit: update endpoint to `mfe_context`.

waheedahmed (commented):
Few minor comments otherwise 👍

PR #253, opened 2021-04-20
add username suggestions functionality

waheedahmed:
Added username suggestions functionality according to redesign.

Backend PR: https://github.com/edx/edx-platform/pull/27387

https://user-images.githubusercontent.com/2851134/115433041-82ad9800-a220-11eb-92b7-f9eb8d6ae4d2.mp4

VAN-52

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -19,7 +19,7 @@ import { ExpandMore } from '@edx/paragon/icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { registerNewUser, fetchRealtimeValidations } from './data/actions';
-import { registrationRequestSelector } from './data/selectors';
+import { registrationRequestSelector, validationsSelector } from './data/selectors';
 import messages from './messages';
 import OptionalFields from './OptionalFields';
 import RegistrationFailure from './RegistrationFailure';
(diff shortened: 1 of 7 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -229,6 +230,21 @@ class RegistrationPage extends React.Component {
...
+    const payload = {
+      username: suggestion,
+    };
+    this.setState({
+      ...payload,
+      fieldName,
+      updateFieldErrors: false,
+      updateAlertErrors: false,
+    }, () => {
+      this.validateInput(fieldName, suggestion, payload, false);
```
zainab-amir, line 244: why do we need to validate the suggestion? is it not already verified before backend sends it to the frontend?

waheedahmed: @zainab-amir It's just to clear out the error message and username suggestions.

src/register/UsernameField.jsx
```diff
@@ -0,0 +1,56 @@
...
+import { usernameSuggestionsSelector } from './data/selectors';
+import { FormGroup } from '../common-components';
+import messages from './messages';
+
+const UsernameField = (props) => {
+  const { intl, usernameSuggestions } = props;
+
+  return (
+    <FormGroup {...props}>
+      {usernameSuggestions && usernameSuggestions.length > 0 ? (
```
zainab-amir, line 17: if the default value is `[]` do we need to add the first check, shouldn't `usernameSuggestions.length` be enough?

zainab-amir (approved):

src/register/UsernameField.jsx
waheedahmed: Yes, it should be.

PR #257, opened 2021-04-23
Redesign login error messages

mubbsharanwar:
Redesign login error messages

  - generic error message appears on top of the page
  - update field related error messages
    
    VAN-396

**Before** 
before

**After**
after

### DIFF
--- a/src/login/LoginFailure.jsx
+++ b/src/login/LoginFailure.jsx
@@ -1,7 +1,8 @@
 import React from 'react';
 
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { Alert } from '@edx/paragon';
+import { Alert, Icon } from '@edx/paragon';
+import { Info } from '@edx/paragon/icons';
 import PropTypes from 'prop-types';
 
 import processLink from '../data/utils';
@@ -80,10 +81,9 @@ const LoginFailureMessage = (props) => {
       break;
     case INVALID_FORM:
       errorList = (
-        <>
-          {context.emailOrUsername && <li key={`${INVALID_FORM}-email`}>{context.emailOrUsername}</li>}
-          {context.password && <li key={`${INVALID_FORM}-password`}>{context.password}</li>}
-        </>
+        <p>
+          {intl.formatMessage(messages['login.form.invalid.error.message'])}
+        </p>
       );
       break;
     case FAILED_LOGIN_ATTEMPT: {
@@ -168,8 +168,9 @@ const LoginFailureMessage = (props) => {
 
   return (
     <Alert id="login-failure-alert" variant="danger">
+      <Icon src={Info} className="alert-icon" />
       <Alert.Heading>{intl.formatMessage(messages['login.failure.header.title'])}</Alert.Heading>
-      <ul>{errorList}</ul>
+      { errorList }
     </Alert>
   );
 };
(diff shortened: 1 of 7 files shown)


src/login/LoginFailure.jsx
```diff
@@ -168,6 +168,7 @@ const LoginFailureMessage = (props) => {
 
   return (
     <Alert id="login-failure-alert" variant="danger">
+      <Icon src={Info} className="alert-icon" />
       <Alert.Heading>{intl.formatMessage(messages['login.failure.header.title'])}</Alert.Heading>
       <ul>{errorList}</ul>
```
zainab-amir, line 173: remove `<ul>`. Use `<p>`. Since there are no unordered list of items we shouldn't use `<ul>`

src/login/messages.jsx
```diff
@@ -198,6 +198,11 @@ const messages = defineMessages({
     defaultMessage: 'To protect your account, it’s been temporarily locked. Try again in {lockedOutPeriod} minutes.',
     description: 'Account locked out user message',
   },
+  'login.form.invalid.error.message': {
+    id: 'login.form.invalid.error.message',
+    defaultMessage: 'Please fill in the fields blow.',
```
zainab-amir, line 203: fix: `below`

src/login/tests/LoginFailure.test.jsx
```diff
@@ -112,7 +111,7 @@ describe('LoginFailureMessage', () => {
       </IntlProvider>,
     );
 
-    const expectedMessage = 'We couldn\'t sign you in.Please enter your email.Please enter your password.';
+    const expectedMessage = 'We couldn\'t sign you in.Please fill in the fields blow.';
```
zainab-amir, line 114: same fix spellings of below

src/login/messages.jsx
```diff
@@ -114,12 +114,12 @@ const messages = defineMessages({
   },
   'email.validation.message': {
     id: 'email.validation.message',
-    defaultMessage: 'Please enter your email.',
+    defaultMessage: 'Enter your username or email.',
```
zainab-amir, line 117: there is not full stop after the message in design

src/login/messages.jsx
```diff
@@ -114,12 +114,12 @@ const messages = defineMessages({
...
     description: 'Validation message that appears when email is empty',
   },
   'password.validation.message': {
     id: 'password.validation.message',
-    defaultMessage: 'Please enter your password.',
+    defaultMessage: 'Enter your password.',
```
zainab-amir, line 122: remove full stop after the message

src/_style.scss
```diff
@@ -456,3 +456,7 @@ select.form-control {
   line-height: 2rem;
   font-weight: 400;
 }
+
+.alert-message {
```
zainab-amir, line 460: No need for custom class, keep the font size as it is. If it needs to be changed, it should be changed in paragon

zainab-amir (commented):
When I switch tabs (login -> register -> login) after an error has occurred, the message doesn't disappear.

src/login/LoginFailure.jsx
mubbsharanwar: In other cases of `errorList` still we are using `li` that's why we should not remove `ul`.
We can change this like `{ errorCode === INVALID_FORM ? <ul>{ errorList }</ul> :  { errorList } }`.

zainab-amir: This is not needed, eventually all error messages will be changed (we are no longer going to use lists for error, see the designs)

mubbsharanwar:
it was merged accidentally without approval.

PR #270, opened 2021-05-05
VAN-408 - Redesign login/register SSO-SAML

shafqatfarhan:


shafqatfarhan:
@waheedahmed Made the suggested changes. PR is ready for review again.

src/login/tests/LoginPage.test.jsx
```diff
@@ -318,6 +307,10 @@ describe('LoginPage', () => {
   });
 
   it('should redirect to social auth provider url on SSO button click', () => {
+    mergeConfig({
```
waheedahmed, line 310: If this is required in most if the test cases, could we move this into `beforeEach()` instead of adding in each test case and disable it for only those where you need to?

src/common-components/Logistration.jsx
```diff
@@ -12,19 +16,47 @@ import { RegistrationPage } from '../register';
 
 const Logistration = (props) => {
   const { intl, selectedPage } = props;
+  const [institutionLogin, setInstitutionLogin] = useState(false);
+
+  const handleInstitutionLogin = () => {
+    if (selectedPage === LOGIN_PAGE) {
```
waheedahmed, line 22: Why we are sending this event only for the login page? I believe we need to send it for both login and register.

src/common-components/Logistration.jsx
```diff
@@ -12,19 +16,47 @@ import { RegistrationPage } from '../register';
...
+      sendTrackEvent('edx.bi.institution_login_form.toggled', { category: 'user-engagement' });
+      sendPageEvent('login_and_registration', 'institution_login');
```
waheedahmed, line 24: And when the user clicks the back button the page event name should be `login` or `register` depending upon where it came from.

shafqatfarhan: I moved this login from the Login component to Logistration but this event tracking was only enabled for Login

src/common-components/Logistration.jsx
shafqatfarhan: Previously, this event tracking was only enabled for the Login screen. I moved this logic from the Login component to Registration. Please let me know if we need to send an event for Registration also

src/login/tests/LoginPage.test.jsx
shafqatfarhan: It is only required in 3 tests out of 24 for LoginPage tests. So, I think moving it to `beforeEach ()` and then disabling it where required would not be a better option.

src/common-components/Logistration.jsx
waheedahmed: Yes, it should be for both login and registration. I guess it was implemented incorrectly in the past..

src/common-components/Logistration.jsx
waheedahmed: I guess there might be some misunderstanding while implemented in the past, you can verify the expected functionality here https://edge.edx.org/register using the `Use my institution/campus credentials` button.

shafqatfarhan: Updated the tracking event

src/common-components/Logistration.jsx
shafqatfarhan: Done

src/common-components/Logistration.jsx
```diff
@@ -12,19 +16,46 @@ import { RegistrationPage } from '../register';
 
 const Logistration = (props) => {
   const { intl, selectedPage } = props;
+  const [institutionLogin, setInstitutionLogin] = useState(false);
+
+  const handleInstitutionLogin = () => {
+    sendTrackEvent('edx.bi.institution_login_form.toggled', { category: 'user-engagement' });
+    sendPageEvent(selectedPage === LOGIN_PAGE ? 'login' : 'register', 'institution_login');
```
waheedahmed, line 23: The category should remain the same `login_and_registration` for all three, only change the name according to the scenario. e.g. `institution_login`, `login`, `register`

shafqatfarhan: @waheedahmed I have updated the logic for the page event name. Please review again

waheedahmed (approved):
Don't forget to squash your commits before merging 🙂

waheedahmed (approved):

PR #279, opened 2021-05-17
Scroll to on top error message

mubbsharanwar:
- Scroll to error message on every submit action of registration page

VAN-478

### DIFF
--- a/src/register/tests/RegistrationPage.test.jsx
+++ b/src/register/tests/RegistrationPage.test.jsx
@@ -191,8 +191,10 @@ describe('RegistrationPage', () => {
       expect(registrationPage.find('div[feedback-for="country"]').text()).toEqual(emptyFieldValidation.country);
 
       const alertBanner = 'We couldn\'t create your account.Please check your responses and try again.';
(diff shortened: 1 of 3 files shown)


src/register/tests/RegistrationPage.test.jsx
```diff
@@ -191,8 +191,10 @@ describe('RegistrationPage', () => {
       expect(registrationPage.find('div[feedback-for="country"]').text()).toEqual(emptyFieldValidation.country);
 
       const alertBanner = 'We couldn\'t create your account.Please check your responses and try again.';
+      registrationPage.find('RegistrationPage').setState({ failureCount: 1 });
```
zainab-amir, line 194: why are we setting the `failureCount` explicitly? Shouldn't this be set automatically when we click the submit button?

zainab-amir (approved):
Looks good, just one question

src/register/tests/RegistrationPage.test.jsx
mubbsharanwar: It should set automatically but not setting up. That's why we need to set `failureCount` explicitly to cover this case in unit test .

PR #282, opened 2021-05-18
update base component

zainab-amir:
Currently the base component resets the form when page size is changed. Moving form to the base component instead of passing it to layouts fixes the issue.

[VAN-461](https://openedx.atlassian.net/browse/VAN-461)


uzairr:
ensure the green build and check the coverage as well,

PR #291, opened 2021-05-21
Fire optimizely event

zainab-amir:
Adding code to track custom events in Optimizely. The event `user_registered_successfully` will be tracked and associated with the current visitor.

Ticket: https://openedx.atlassian.net/browse/VAN-487


waheedahmed (approved):
Will it complete before the redirection actually happens after successful registration?

PR #298, opened 2021-05-26
Clear field error on focus

mubbsharanwar:
- Clear field related error messages for all forms on focus in event

VAN-505

**Way to test**

- Submit empty form or focus out input field, field related error message will be appear.
- On focus in event field related error message will be disappear.


src/common-components/AuthnValidationFormGroup.jsx
```diff
@@ -7,7 +7,9 @@ import {
 } from '@edx/paragon';
 
 const AuthnCustomValidationFormGroup = (props) => {
-  const { onClick, onChange, onBlur } = props;
+  const {
```
zainab-amir, line 10: In redesign we are not using this component. We are using FormGroup

zainab-amir (changes requested):

src/common-components/AuthnValidationFormGroup.jsx
waheedahmed: Can we remove this component if it's not in use?

waheedahmed: @mubbsharanwar I guess it's currently in use in https://github.com/edx/frontend-app-authn/blob/d2b05d770c72f363e999f791eeca5a9b3d1d1a1f/src/reset-password/ResetPasswordPage.jsx#L156 and will get removed with the redesign work of that page. You can skip it for now.

src/common-components/PasswordField.jsx
```diff
@@ -21,6 +21,13 @@ const PasswordField = (props) => {
     setShowTooltip(props.showRequirements && false);
   };
 
+  const handleFocus = (e) => {
+    if (props.handleFocus) {
+      props.handleFocus(e);
+      setTimeout(() => setShowTooltip(props.showRequirements && true), 150);
+    }
+  };
```
zainab-amir, line 29: setTimeout should be called even if handleFocus is not present
```suggestion
  const handleFocus = (e) => {
    if (props.handleFocus) {
      props.handleFocus(e);
    }
    setTimeout(() => setShowTooltip(props.showRequirements && true), 150);
  };
```

src/login/LoginPage.jsx
```diff
@@ -91,6 +91,16 @@ class LoginPage extends React.Component {
     this.props.loginRequest(payload);
   }
 
+  handleOnFocus = (e) => {
+    const { errors } = this.state;
+    if (e.target.name === 'email') {
+      errors.emailOrUsername = '';
+    } else {
+      errors[e.target.name] = '';
+    }
```
zainab-amir, line 100: can't this be written as (we can change the email field name):
```suggestion
   errors[e.target.name] = '';
```

src/reset-password/tests/ResetPasswordPage.test.jsx
```diff
@@ -167,6 +167,21 @@ describe('ResetPasswordPage', () => {
...
+      token_status: 'valid',
+      token: 'token',
+      errors: '',
+    };
+    const resetPasswordPage = mount(reduxWrapper(<IntlResetPasswordPage {...props} />));
+    resetPasswordPage.find('button.btn-primary').simulate('click');
+
+    resetPasswordPage.update();
+    resetPasswordPage.find('input#reset-password-input').simulate('focus');
+    resetPasswordPage.find('input#confirm-password-input').simulate('focus');
```
zainab-amir, line 182: shouldn't you check if field error is removed after these statements?

zainab-amir (changes requested):

zainab-amir (approved):

PR #300, opened 2021-05-27
VAN-506 - Remove icon in form validation messages

shafqatfarhan:


waheedahmed (approved):
The current version of the `frontend-platform` supports Paragon version < 14.0.0, do we need to update that as well?
```
npm WARN @edx/frontend-platform@1.8.4 requires a peer of @edx/paragon@>=10.0.0 <14.0.0 but none is installed. You must install peer dependencies yourself.
```

Otherwise looks good 👍

PR #325, opened 2021-06-09
alert styling updated

adeelehsan:
Screen Shot 2021-05-31 at 7 02 57 PM
VAN-499

### DIFF
--- a/src/common-components/AlertDismissible.jsx
+++ b/src/common-components/AlertDismissible.jsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Alert, Icon } from '@edx/paragon';
+import { Alert } from '@edx/paragon';
 import { Info } from '@edx/paragon/icons';
 import { injectIntl } from '@edx/frontend-platform/i18n';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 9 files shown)


src/common-components/AlertDismissible.jsx
```diff
@@ -8,8 +8,7 @@ function AlertDismissible(props) {
   const [show, setShow] = useState(true);
 
   return show ? (
-    <Alert variant={props.variant} onClose={() => setShow(false)} dismissible className="pb-1 pt-1">
-      <Icon src={Info} className="alert-icon" />
+    <Alert variant={props.variant} onClose={() => setShow(false)} dismissible className="pb-1 pt-1" icon={Info}>
```
mubbsharanwar, line 11: use `Error ` instead of `Info` icon like using in other components.

mubbsharanwar (approved):
It looks good just minor change required.

src/common-components/AlertDismissible.jsx
adeelehsan: I think variant may be different for different components that's why we are using `Info`

PR #327, opened 2021-06-09
Password show and hide icon replaced with paragon icon

adeelehsan:
VAN-523


mubbsharanwar:
`Visibility` and  `VisibilityOff` Icon shows only on mouse hover. These should show  without hover.

Icon missing 
Screen Shot 2021-06-09 at 12 47 58 PM

Icon show when mouse hover on icon
Screen Shot 2021-06-09 at 12 48 04 PM

Also icon is not according to figma design.

adeelehsan:
working find at my end.
https://user-images.githubusercontent.com/8156837/121316733-0cc1d500-c923-11eb-98be-33f6204aaf56.mov

mubbsharanwar (approved):

PR #331, opened 2021-06-11
fix: Logistration tabs behavior

mubbsharanwar:
- Logistration Tabs behavior and styles fix by using Paragon Tab component.

VAN-554

### DIFF
--- a/src/_style.scss
+++ b/src/_style.scss
@@ -35,21 +35,11 @@ $accent-a-light: #c9f2f5;
 .nav-tabs {
   a {
     color: $primary !important;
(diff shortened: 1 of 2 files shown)


src/_style.scss
```diff
@@ -32,6 +32,12 @@ $accent-a-light: #c9f2f5;
   margin-bottom: 0.75rem;
 }
 
+.tab-background {
```
waheedahmed, line 35: No need to define a new class, move `a:hover` styles inside `.nav-tabs`.

waheedahmed (approved):

PR #333, opened 2021-06-11
fix: fix email validation and events

waheedahmed:
Fixed email validation logic and clear error state on close. Also fixed analytic events and removed unused components.

VAN-538, VAN-568


shafqatfarhan (approved):
LGTM. Just a small nit: Would not it be better to place all the implementation related to form validation/errors in the FormGroup component?

waheedahmed:
@shafqatfarhan That's a good suggestion can definitely look into it in future but we won't be able to make those right now.

PR #347, opened 2021-06-15
VAN-613 - Updated styles for Need help signing in? button

shafqatfarhan:


src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -109,8 +110,15 @@ const ForgotPasswordPage = (props) => {
                   onClick={handleSubmit}
                   onMouseDown={(e) => e.preventDefault()}
                 />
-                <Hyperlink id="forgot-password" className="btn btn-link font-weight-500 text-body" destination={getConfig().LOGIN_ISSUE_SUPPORT_LINK}>
-                  {intl.formatMessage(messages['need.help.sign.in.text'])}
+                <Hyperlink
+                  id="forgot-password"
+                  className="btn btn-link font-weight-500 text-body"
```
waheedahmed, line 115: Remove `btn btn-link` and add `ml-4` fixes the underline issue. You can remove the custom styles with this.

waheedahmed (approved):
One minor suggestion otherwise looks good.

src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -109,8 +110,15 @@ const ForgotPasswordPage = (props) => {
...
+                  destination={supportUrl}
+                  onClick={e => {
```
waheedahmed, line 117: Can we use `target="_blank"` instead?

src/forgot-password/ForgotPasswordPage.jsx
shafqatfarhan: sounds good. Will push the change

shafqatfarhan: done

PR #350, opened 2021-06-16
add help text for fullname field

zainab-amir:
#### Context
- Added help text for full name field on registration page (see https://openedx.atlassian.net/browse/VAN-542)
- Updated paragon package to incorporate `.micro` class line height changes (see https://openedx.atlassian.net/browse/VAN-546)
- Added `.text-muted` to disclaimer text after feedback from UX team

### DIFF
--- a/src/register/messages.jsx
+++ b/src/register/messages.jsx
@@ -33,6 +33,11 @@ const messages = defineMessages({
     description: 'Placeholder for the country options dropdown.',
   },
   // Help text
+  'help.text.name': {
+    id: 'help.text.name',
+    defaultMessage: 'This name will be used by any certificates that you earn.',
+    description: 'Help text for fullname field on registration page',
+  },
   'help.text.username.1': {
     id: 'help.text.username.1',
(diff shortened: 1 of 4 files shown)


src/register/messages.jsx
```diff
@@ -33,6 +33,11 @@ const messages = defineMessages({
     description: 'Placeholder for the country options dropdown.',
   },
   // Help text
+  'help.text.name': {
+    id: 'help.text.name',
+    defaultMessage: 'This name will be used by any certificates that you earn.',
```
shafqatfarhan, line 38: The original message on the ticket does not have a full stop (.) at the end `This name will be used by any certificates that you earn`

shafqatfarhan (approved):

src/register/messages.jsx
zainab-amir: yeah I noticed that but other help text messages do have full stop that is why I have added it. Will merge and move this to review and see if Jon needs me to update it.

PR #354, opened 2021-06-16
fix typos in login failure messages

zainab-amir:
- Fixed a few typos in the login failure messages
- Updated messages to be inside paragraphs instead of being separated by line breaks. 

Screenshot 2021-06-16 at 6 45 50 PM
Screenshot 2021-06-16 at 6 46 01 PM

Ticket: https://openedx.atlassian.net/browse/VAN-612


mubbsharanwar:
image-20210612-050159

This error message is not reproducing after several attempts.

mubbsharanwar (approved):

PR #362, opened 2021-06-17
fix anomalies in dropdown

uzairr:
Fix anomalies like:
1. spacing between the borderline and dropdown
2. icon alignment
3. Country list on icon click.

VAN-559

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -622,7 +622,7 @@ select.form-control {
 }
 
 .-mt-4 {
-    margin-top: -1.5rem;
+    margin-top: -1.4rem;
 }
 
 .alert .alert-icon {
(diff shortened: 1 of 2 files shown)


mubbsharanwar (approved):

src/redesign/_style.scss
```diff
@@ -667,3 +667,7 @@ select.form-control {
   max-height: 1rem;
   margin-top:6px;
 }
+
+.pgn__icon {
```
adeelehsan, line 671: That would change the width of all the icons. Make it specific to that icon.

adeelehsan (changes requested):

src/redesign/_style.scss
uzairr: changes addressed!

PR #363, opened 2021-06-18
fix: Password field icon size

mubbsharanwar:
- set password field icon size according to figma design.

**Before:**
before
**After:**
after

VAN-523

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -648,13 +648,6 @@ select.form-control {
       text-decoration: underline;
     }
 }
-.password-visibility{
-  .btn-icon__icon-container .btn-icon__icon {
-    height: 1.5rem !important;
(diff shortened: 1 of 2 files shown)


src/redesign/_style.scss
```diff
@@ -646,7 +646,10 @@ select.form-control {
     }
 }
 .password-visibility{
-  .btn-icon__icon-container .btn-icon__icon {
+  height: 1.5rem !important;
+  width: 1.5rem !important;
+  color: $gray-500;
+  &:hover {
```
adeelehsan, line 652: At hover password icon would be white and icon button should be gray. Can you please double check it again? Its only setting gray color on hover.

adeelehsan (changes requested):

waheedahmed (approved):

PR #364, opened 2021-06-21
fix: Email suggestion styles

mubbsharanwar:
- email suggestion style fix according to figma style.

VAN-601

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -624,25 +624,45 @@ select.form-control {
   }
 }
 
-.alert .alert-icon {
-  margin-top: 3px;
-}
-.alert-dismissible .close {
-  top: 4px;
-  font-weight: 500;
+.email-error-alert {
+  padding: 0.5rem 1rem;
 
-  :hover {
-      font-weight: 700;
+  .alert-icon {
+    margin-top: 4px;
+  }
+
+  .alert-close {
+    float: right;
+    margin-top: 3px;
+
+    &:hover {
+      cursor: pointer;
     }
+  }
+
+  .alert-link {
+    color: $primary !important;
+
+    &:hover {
+      text-decoration: underline;
+      color: $info-700 !important;
+    }
+  }
 }
-.email-top-domain-suggestion-alert-link {
-  color: $primary !important;
+
+.email-warning-alert-link {
+  color: $info-500 !important;
+
+  &:hover {
+    text-decoration: underline;
+    color: $info-700 !important;
+  }
 }
 
 #forgot-password {
- &:hover {
-      text-decoration: underline;
-    }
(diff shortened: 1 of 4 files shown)


src/redesign/register/RegistrationPage.jsx
```diff
@@ -400,7 +400,7 @@ class RegistrationPage extends React.Component {
               onClick={e => { this.handleSuggestionClick(e, this.state.emailErrorSuggestion); }}
             >
               {this.state.emailErrorSuggestion}
-            </Alert.Link>?
+            </Alert.Link>?<Icon src={Close} className="alert-close" onClick={this.handleOnClose} tabIndex="0" />
```
waheedahmed, line 403: I guess it should be an `IconButton` otherwise, nothing will happen on hover.

waheedahmed (changes requested):
The link color is still not according to Figma, also update the Paragon version in this same PR to test all these changes.
Screenshot 2021-06-28 at 12 31 38 PM

src/redesign/register/RegistrationPage.jsx
mubbsharanwar: Using `IconButton` background color will display on hover. We are using `alert-close` custom class in which `cursor: pointer`.

src/redesign/_style.scss
```diff
@@ -632,24 +632,40 @@ select.form-control {
 }
 
 .alert .alert-icon {
```
waheedahmed, line 634: This shouldn't apply to all alert icons and instead, only apply to the email alert icon. For example, check the error icon in the following screenshot which is not aligned with the header text of the alert due to this:
Screenshot 2021-06-29 at 2 31 06 PM

src/redesign/_style.scss
```diff
@@ -632,24 +632,40 @@ select.form-control {
...
+  margin-top: 4px;
 }
-.alert-dismissible .close {
-  top: 4px;
-  font-weight: 500;
 
-  :hover {
-      font-weight: 700;
-    }
+.alert-close {
```
waheedahmed, line 638: Same here, make it email alert specific.

src/redesign/_style.scss
```diff
@@ -631,25 +631,45 @@ select.form-control {
...
-  margin-top: 3px;
-}
-.alert-dismissible .close {
-  top: 4px;
-  font-weight: 500;
+.email-error-alert {
+  padding: 0.5rem 1rem;
+
+  .alert-icon {
+    margin-top: 3px;
```
waheedahmed, line 638:
```suggestion
    margin-top: 4px;
```

waheedahmed (approved):
One small nit otherwise 👍

PR #365, opened 2021-06-23
UI issues fixed

adeelehsan:
VAN 624 & 629

### DIFF
--- a/src/redesign/login/LoginFailure.jsx
+++ b/src/redesign/login/LoginFailure.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { Alert } from '@edx/paragon';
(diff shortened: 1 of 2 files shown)


src/redesign/login/LoginFailure.jsx
```diff
@@ -73,7 +73,7 @@ const LoginFailureMessage = (props) => {
       break;
     case FAILED_LOGIN_ATTEMPT: {
       resetLink = (
-        <a href="/reset">
+        <a href="/reset" style={{ 'text-decoration': 'underline' }}>
```
shafqatfarhan, line 76: nit: We can use [Hyperlink](use https://paragon-edx.netlify.app/components/hyperlink/) here (and wherever required) with its prop `isInline` set to true

shafqatfarhan (commented):
One thing I noticed that the font size for the input fields is not 14px as per Figma.

uzairr (approved):

PR #367, opened 2021-06-24
fix a11y in dropdown

uzairr:
fix a11y related concerns in country dropdown.

VAN-559

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -528,12 +528,11 @@ select.form-control {
 }
 
 .username-suggestion {
-  padding: 0.15rem 0.5rem;
+  padding: 2px 0.5rem;
   margin: 0.25rem;
   border-radius: 0.5rem;
-  line-height: 2rem;
-  font-weight: 400;
-  font-size: 0.875rem;
+  line-height: 24px;
+  font-size: 12px;
 }
 
 .username-suggestion-label {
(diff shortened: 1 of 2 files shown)


src/redesign/register/CountryDropdown.jsx
```diff
@@ -88,17 +88,9 @@ class CountryDropdown extends React.Component {
...
-    }
-    this.setState({
-      dropDownItems, icon: ExpandLess, errorMessage: '', showFieldError: false,
-    });
-
-    if (this.state.dropDownItems?.length > 0) {
-      this.setState({ dropDownItems: '', icon: ExpandMore });
+    const dropDownItems = this.getItems(e.target.value);
+    if (dropDownItems?.length > 1) {
+      this.setState({ dropDownItems, icon: this.expandMoreButton() });
```
waheedahmed, line 93: Need to set `errorMessage: ''` to clear the error message on `onFoucs`.

waheedahmed (commented):
Also, we need to validate the value as well, currently we are just checking if the value is set which isn’t True with the typeahead search field because if the user enters an incorrect value; it does not show an error message on focus out.
Screenshot 2021-06-24 at 2 29 57 PM

src/redesign/_style.scss
```diff
@@ -528,7 +528,7 @@ select.form-control {
 }
 
 .username-suggestion {
-  padding: 0.15rem 0.5rem;
```
waheedahmed, line 531: This is not the desired change, according to Figma design `padding-left/right` should be 2px and `font-size: 12px` and `line-height: 24px`.

uzairr: Screen Shot 2021-06-24 at 5 52 18 PM
I think this is not promissing with design ?

waheedahmed (approved):

PR #378, opened 2021-06-30
fix: username suggestion chip styles

waheedahmed:
VAN-609

## Figma
Screenshot 2021-06-30 at 12 14 45 PM

## Before
Screenshot 2021-06-30 at 12 15 54 PM

## After
Screenshot 2021-06-30 at 1 50 40 PM

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -532,15 +532,18 @@ select.form-control {
 }
 
 .username-suggestion {
-  padding: 2px 0.5rem;
+  padding: 1px 0.5rem;
   margin: 0.25rem;
-  border-radius: 0.5rem;
+  border-radius: 0.375rem;
   line-height: 24px;
   font-size: 12px;
+  font-weight: normal;
+  color: $primary-700;
 }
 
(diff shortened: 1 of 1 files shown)


src/redesign/_style.scss
```diff
@@ -532,15 +532,17 @@ select.form-control {
 }
 
 .username-suggestion {
-  padding: 2px 0.5rem;
+  padding: 1px 0.5rem;
```
shafqatfarhan, line 535: nit: Might need to update padding-top to 2px as Figma shows padding `padding: 2px 8px;`.

shafqatfarhan (approved):

shafqatfarhan:
@waheedahmed Suggested button color is `Primary/700`

mubbsharanwar (approved):

waheedahmed:
@shafqatfarhan nice catch, done!

src/redesign/_style.scss
waheedahmed: @shafqatfarhan it increases the height to 30px but Figma suggests 28px.

PR #380, opened 2021-07-01
feat: set register conversion cookie

waheedahmed:
Set register conversion cookie upon successful registration to fire social conversion pixels in GTM.


shafqatfarhan (approved):
Changes look good. A question: do we not need to delete the cookie as specified in ticket's description?

waheedahmed:
@shafqatfarhan We will delete it in GTM after firing the required tags.

PR #383, opened 2021-07-02
fix: fix overlapping over logo

mubbsharanwar:
- fix start learning with edx text overlapping over logo.

**Before**
Screen Shot 2021-07-02 at 3 30 37 PM

**After**
after

VAN-606

### DIFF
--- a/src/redesign/_style.scss
+++ b/src/redesign/_style.scss
@@ -709,3 +709,9 @@ select.form-control {
   margin-top: 0.188rem !important;
   line-height: 1.25rem;
 }
+
+@media (max-height: 500px) {
+  .large-screen-svg-line, .large-heading {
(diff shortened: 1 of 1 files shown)


shafqatfarhan (approved):
Looks good. One small nit:

https://github.com/edx/frontend-app-authn/blob/redesign-master/src/redesign/common-components/LargeScreenLeftLayout.jsx#L12
`<div className="min-vh-100 pr-0 mt-lg-n2 d-flex align-items-center">` can be replaced with 
`<div className="min-vh-100 pr-0 mt-3 pb-3 d-flex align-items-center">`

by adding `mt-3 pb-3` instead of `mt-lg-n2`

waheedahmed (approved):

PR #405, opened 2021-07-29
chore: add welcome page events

zainab-amir:
#### context:
It would be nice to know with this AB, out of all registered users
- how many submit demo data
- how many skip welcome page
- how many exit our site from welcome page (`authn_pp_conversion` is used for this calculation)

Ticket: https://openedx.atlassian.net/browse/VAN-676


zainab-amir:
The same events have been captured directly via optimizely's `click events`. Closing this as it is not required anymore.

waheedahmed (approved):

PR #409, opened 2021-08-03
fix: Side banner didn't show.

attiyaIshaque:


src/common-components/Logistration.jsx
```diff
@@ -16,7 +16,9 @@ import { LOGIN_PAGE, REGISTER_PAGE } from '../data/constants';
 import { updatePathWithQueryParams, getTpaHint } from '../data/utils';
 import { LoginPage } from '../login';
 import { RegistrationPage } from '../register';
+import BaseComponent from '../base-component';
 
+jest.mock('@edx/frontend-platform/auth');
```
waheedahmed, line 21: I guess this is added by mistake and needs to be removed?

waheedahmed (changes requested):
Also need to update `src/welcome/WelcomePage.jsx`.

src/common-components/Logistration.jsx
attiyaIshaque: Yup. Copy/paste in wrong file. I will removed it. Thanks for pointing this.

PR #410, opened 2021-08-04
refactor: enable welcome page for all users

zainab-amir:
Currently welcome page is behind an optimizely experiment flag. Removing it will enable it for all users.

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -502,13 +502,6 @@ class RegistrationPage extends React.Component {
           value: this.state.totalRegistrationTime,
         },
       });
-
-      if (window.optimizelyExperimentName !== 'release-pp-concept-1') {
-        window.optimizely.push({
-          type: 'event',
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -502,13 +502,6 @@ class RegistrationPage extends React.Component {
           value: this.state.totalRegistrationTime,
         },
       });
-
-      if (window.optimizelyExperimentName !== 'release-pp-concept-1') {
-        window.optimizely.push({
-          type: 'event',
-          eventName: 'authn_pp_conversion',
```
waheedahmed, line 509: Do we need to remove this event from welcome page on submit/skip click as well?

waheedahmed (approved):
Just one question otherwise 👍

src/register/RegistrationPage.jsx
zainab-amir: yeah will remove it

PR #412, opened 2021-08-09
fix: force refresh JWT to fix redirct loop

waheedahmed:
User trapped into a redirect loop if backend session expires but has a valid JWT, fixed by force refresh JWT token.

VAN-682


attiyaIshaque (approved):
👍

PR #416, opened 2021-08-10
fix: [VAN-332] Full name validation on registration page.

attiyaIshaque:


waheedahmed (commented):
Two minor nits otherwise looks good.

waheedahmed (approved):

PR #422, opened 2021-08-17
chore: send analytic events for Welcome page

waheedahmed:
VAN-679


zainab-amir (approved):
Looks good, do we need to remove the optional field properties from the register event at the backend?

waheedahmed:
@zainab-amir I guess no, there might be some OpenedX user is using them.

PR #426, opened 2021-08-31
chore: add 15% discount incentive to increase registrations.

attiyaIshaque:
## https://openedx.atlassian.net/browse/VAN-687
Add experiment detail of 15% discount incentive to increase registrations. Value will be based on the variant sent by optimizely.
Three Variation we have used. 
- default
- Variation 1
- Variation 2
with ~33% traffic divided into all 3 types.


zainab-amir:
| variation 1 | variation 2|
|-----|-----|
Screenshot 2021-09-01 at 2 55 12 PM|Screenshot 2021-09-01 at 2 55 25 PM

waheedahmed (approved):

PR #444, opened 2021-09-29
fix: VAN-719 - Handled password field error violation on username change

shafqatfarhan:


mubbsharanwar (approved):

mubbsharanwar (approved):
Looks good. Just need to add unit test of this patch.

PR #449, opened 2021-10-12
feat: [VAN-655] Put User's year of birth behind the feature flag

attiyaIshaque:
## Description

As a B2C web app user, when I register with edX, I am not asked for the year of birth information on the Welcome page of the onboarding step.
Remove the ‘year of birth’ field from the Welcome page on Authn MFE if feature flag is disabled.

## Supporting information

Ticket: https://openedx.atlassian.net/browse/VAN-655

## Testing instructions

Fixed unit test


src/welcome/WelcomePage.jsx
```diff
@@ -170,19 +170,22 @@ const WelcomePage = (props) => {
...
-                value={values.yearOfBirth}
-                onChange={(e) => onChangeHandler(e)}
-                trailingElement={<Icon src={ExpandMore} />}
-                floatingLabel={intl.formatMessage(messages['year.of.birth.label'])}
-              >
-                <option value="">{intl.formatMessage(messages['year.of.birth.label'])}</option>
-                {getOptions('yearOfBirth')}
-              </Form.Control>
-            </Form.Group>
+            {(getConfig().COLLECT_YEAR_OF_BIRTH === 'false')
```
waheedahmed, line 173: I believe it should be:
```suggestion
            {(getConfig().COLLECT_YEAR_OF_BIRTH === 'true')
```

.env
```diff
@@ -25,4 +25,4 @@ DISABLE_ENTERPRISE_LOGIN=''
 REGISTER_CONVERSION_COOKIE_NAME=null
 ENABLE_PROGRESSIVE_PROFILING=''
 MARKETING_EMAILS_OPT_IN='false'
-COLLECT_YEAR_OF_BIRTH='true'
```
waheedahmed, line 28: Why did we remove the value here?

waheedahmed: I believe the default is always true, we will disable it through the remote config.

waheedahmed: Those were for different purposes and an empty string makes sense in those cases, but in our case, we need to keep the field enabled in default behavior and only disable it for edX and which should be done through the remote config.

PR #456, opened 2021-10-27
feat: fire track event

mubbsharanwar:
- fire track event for save for latter course.
- event will fire when user click on enroll button in email template.

**This PR depends on:**
https://github.com/edx/edx-platform/pull/29089

VAN-741

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -96,6 +96,10 @@ class RegistrationPage extends React.Component {
 
     const payload = { ...this.queryParams };
 
+    if (payload.save_for_later === 'true') {
+      sendTrackEvent('edx.bi.user.save.course.enroll.clicked', { category: 'save-for-later' });
+    }
+
     if (this.tpaHint) {
       payload.tpa_hint = this.tpaHint;
     }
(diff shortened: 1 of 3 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -94,6 +94,10 @@ class RegistrationPage extends React.Component {
 
     const payload = { ...this.queryParams };
 
+    if (payload.save_for_later === 'true') {
+      sendTrackEvent('edx.user.enroll.favourite.course', { category: 'save-for-later' });
```
waheedahmed, line 98: nit: `edx.bi.user.save.course.enroll.clicked`

waheedahmed (approved):
One suggestion otherwise 👍

waheedahmed (commented):
If possible please add a test case if `save_for_later` is present the triggers the `sendTrackEvent` event with the required arguments.

src/register/tests/RegistrationPage.test.jsx
```diff
@@ -703,6 +703,14 @@ describe('RegistrationPage', () => {
       expect(analytics.sendPageEvent).toHaveBeenCalledWith('login_and_registration', 'register');
     });
 
+    it('should send track event for save_for_later parm', () => {
```
waheedahmed, line 706: type: param

PR #462, opened 2021-11-03
build: use shared browserslist config

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.


waheedahmed (approved):
One question, since older browser support is dropped, can we remove this rule https://github.com/edx/frontend-app-authn/blob/007dfa82a8a7423c8c438d2d36731986a11a7123/webpack.prod.config.js#L5 to stop transforming `fastest-levenshtein` to ES5?

UsamaSadiq (approved):

zainab-amir:
@adamstankiewicz I just noticed while going through open PRs that you created this a long time ago. I did the same thing and removed the es-5 check https://github.com/openedx/frontend-app-authn/pull/660.

PR #465, opened 2021-11-10
chore: remove opt-in/opt-out experiment code

waheedahmed:
Removed opt-in/opt-out experiment code to enable it for all users using feature flag only.


zainab-amir (approved):
Since this is not a part of experiment we should add a test for it.

waheedahmed:
@zainab-amir Test case added, please take another look.

PR #469, opened 2021-11-16
fix: remove aol domain from email provider list

attiyaIshaque:
### Description
Remove the Confusion for any new users who still have an msn.com email address.

### Before fix:
Screenshot 2021-11-16 at 6 32 21 PM

### After Fix:
Screenshot 2021-11-16 at 6 30 15 PM


shafqatfarhan (approved):
👍

PR #471, opened 2021-12-03
feat: VAN-666 - Reject new password that is detected as vulnerable

shafqatfarhan:
[VAN-666](https://openedx.atlassian.net/browse/VAN-666)

### DIFF
--- a/src/reset-password/ResetPasswordPage.jsx
+++ b/src/reset-password/ResetPasswordPage.jsx
@@ -53,7 +53,11 @@ const ResetPasswordPage = (props) => {
   const validatePasswordFromBackend = async (password) => {
     let errorMessage = '';
     try {
-      errorMessage = await validatePassword(password);
+      const payload = {
+        reset_password_page: true,
+        password,
+      };
+      errorMessage = await validatePassword(payload);
     } catch (err) {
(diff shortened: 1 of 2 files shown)


src/reset-password/ResetPasswordPage.jsx
```diff
@@ -53,7 +53,11 @@ const ResetPasswordPage = (props) => {
   const validatePasswordFromBackend = async (password) => {
     let errorMessage = '';
     try {
-      errorMessage = await validatePassword(password);
+      const payload = {
+        reset_password_page: true,
```
waheedahmed, line 57: Do we still need this since we have a separate feature flag at the backend?

waheedahmed (approved):
One question otherwise 👍

src/reset-password/ResetPasswordPage.jsx
shafqatfarhan: yes, this is required along with the feature flag to enable validation only for reset password page

shafqatfarhan: We are using the same Validation API endpoint for registration and sign-in validations as well

PR #504, opened 2022-01-10
perf: csrf token call

mubbsharanwar:
do options call on page load time for login api to reduce request response time.
VAN-715

### DIFF
--- a/src/common-components/Logistration.jsx
+++ b/src/common-components/Logistration.jsx
@@ -4,6 +4,8 @@ import { Redirect } from 'react-router-dom';
 
 import { sendPageEvent, sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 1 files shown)


src/common-components/Logistration.jsx
```diff
@@ -33,6 +36,13 @@ const Logistration = (props) => {
     if (renameRegisterExperiment) {
       setRegisterRenameExpVariation(renameRegisterExperiment);
     }
+    if (!csrfTokenCall) {
```
waheedahmed, line 39: I don't think we need to manage a state for this since hook will run only once. And, CSRF service also maintains a cache for the token so if somehow a second call is initiated it will get the cached token and won't make a second call to the backend.

waheedahmed (changes requested):

src/common-components/Logistration.jsx
mubbsharanwar: Remove local state, working fine. Thanks for suggestion.

waheedahmed (approved):

PR #507, opened 2022-01-12
feat: improve register response time

zainab-amir:
In order to decrease time in between when user clicks the `Create account` button and successfully get registered;
- removed validation call from the submission function
- ~~did the OPTIONS call at the time of page load~~

[VAN-711](https://openedx.atlassian.net/browse/VAN-711)


waheedahmed (approved):
Looks great! 👍

PR #528, opened 2022-02-04
feat: remove optional fields from register form

zainab-amir:
This is part of the larger ticket where we will move optional fields to progressive profiling page.

[VAN-837](https://openedx.atlassian.net/browse/VAN-837)


.env.development
```diff
@@ -23,7 +23,6 @@ AUTHN_MINIMAL_HEADER=true
 LOGIN_ISSUE_SUPPORT_LINK='/login-issue-support-url'
 TOS_AND_HONOR_CODE='http://localhost:18000/honor'
 PRIVACY_POLICY='http://localhost:18000/privacy'
-REGISTRATION_OPTIONAL_FIELDS=''
```
waheedahmed, line 26: Are we defining a new variable for optional fields on the progressive profiling page?

src/register/OptionalFields.jsx
```diff
@@ -1,110 +0,0 @@
...
-
-import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { getConfig } from '@edx/frontend-platform';
-import { Form, Icon } from '@edx/paragon';
-import { ExpandMore } from '@edx/paragon/icons';
-
-import { EDUCATION_LEVELS, GENDER_OPTIONS, YEAR_OF_BIRTH_OPTIONS } from './data/constants';
-import messages from './messages';
-
-const OptionalFields = (props) => {
```
waheedahmed, line 12: Instead of removing, can we utilize these on the progressive profiling page in some way?

waheedahmed (approved):
Couple of questions, otherwise 👍

.env.development
zainab-amir: none yet. We might not need it.

src/register/OptionalFields.jsx
zainab-amir: We will probably not need this code for our approach but the reason for creating this PR was to remove the fields from Registration folder. These are already duplicated in progressive profiling folder.

PR #536, opened 2022-03-08
chore: update transifex scripts to api v3

waheedahmed:
As of April 7th, 2022 the API versions 2 and 2.5 will no longer be operational and relevant requests will begin to fail. Switched to `@edx/reactifex` from `reactifex` which does support V3.

VAN-894

### DIFF
--- a/src/login/messages.jsx
+++ b/src/login/messages.jsx
@@ -20,7 +20,7 @@ const messages = defineMessages({
   'sign.in.button': {
     id: 'sign.in.button',
     defaultMessage: 'Sign in',
-    description: 'Button label that appears on login page',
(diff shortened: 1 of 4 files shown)


src/login/messages.jsx
```diff
@@ -20,7 +20,7 @@ const messages = defineMessages({
   'sign.in.button': {
     id: 'sign.in.button',
     defaultMessage: 'Sign in',
-    description: 'Button label that appears on login page',
+    description: 'Sign in button label that appears on login page',
```
waheedahmed, line 23: Changed to test the push translation job.

shafqatfarhan (approved):
LGTM 👍

PR #537, opened 2022-03-09
feat: VAN-876 - Rename register experiment with improved metrics

shafqatfarhan:
https://openedx.atlassian.net/browse/VAN-876

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import Skeleton from 'react-loading-skeleton';
 import { Helmet } from 'react-helmet';
 import PropTypes, { string } from 'prop-types';
+import classNames from 'classnames';
 
 import { getConfig } from '@edx/frontend-platform';
 import { sendPageEvent, sendTrackEvent } from '@edx/frontend-platform/analytics';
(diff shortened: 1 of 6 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -72,24 +73,31 @@ class RegistrationPage extends React.Component {
...
   }
 
   componentDidMount() {
     window.optimizely = window.optimizely || [];
-    window.optimizely.push({
-      type: 'page',
-      pageName: 'authn_registration_page',
-      isActive: true,
-    });
-
```
zainab-amir, line 90: Lets keep this as it is. We don't need to move it behind a condition. If we don't need it for this experiment we can skip it in Optimizely metric.

src/register/RegistrationPage.jsx
```diff
@@ -72,24 +73,31 @@ class RegistrationPage extends React.Component {
...
     const payload = { ...this.queryParams };
 
+    if (payload.register_for_free === 'true') {
+      window.optimizely.push({
+        type: 'event',
+        eventName: 'authn-registration-page',
```
zainab-amir, line 91:
```suggestion
        eventName: 'van-876-authn-registration-page',
```

zainab-amir (changes requested):

src/register/RegistrationPage.jsx
shafqatfarhan: done

zainab-amir (approved):

PR #567, opened 2022-04-26
fix: Update position of cross icon on username field

syedsajjadkazmii:
The Cross icon was overlapping the border of the username field when username suggestions were generated,
This PR fixes the position of the cross icon of username suggestions field

[VAN-936](https://openedx.atlassian.net/browse/VAN-936)

**Before**

Screen Shot 2022-04-22 at 2 48 45 PM

**After**

Screen Shot 2022-04-26 at 1 22 35 PM

### DIFF
--- a/src/_style.scss
+++ b/src/_style.scss
@@ -866,7 +866,7 @@ select.form-control {
   margin-left: 15px;
 }
 .suggested-username-close-button {
-  right: 0;
+  right: 1rem;
   position: absolute;
 }
 .suggested-username-with-error {


src/_style.scss
```diff
@@ -866,7 +866,7 @@ select.form-control {
   margin-left: 15px;
 }
 .suggested-username-close-button {
-  right: 0;
+  right: 3%;
```
shafqatfarhan, line 869: alternatively, we could use 1rem

shafqatfarhan (approved):
Just a small nit, otherwise LGTM

src/_style.scss
syedsajjadkazmii: yes you are right, thankyou

PR #569, opened 2022-04-28
fix: Styling issue in fields trailing element

syedsajjadkazmii:
The trailing element in a field is out of position in some browsers that do not support `inset-inline-end` property such as Safari version 13 and less. This PR fixes this problem for all browsers

**Before**

Screen Shot 2022-04-28 at 6 26 24 PM

**After**

Screen Shot 2022-04-28 at 6 26 46 PM


shafqatfarhan (approved):
👍

PR #571, opened 2022-05-09
fix: Add aria-Invalid="true" to error-state fields

syedsajjadkazmii:
Added aria-Invalid="true" to error-state fields in /login, /register and /reset pages

[VAN-548](https://2u-internal.atlassian.net/browse/VAN-548)


shafqatfarhan (approved):
LGTM 👍

PR #574, opened 2022-05-17
docs: ADR to enable 2FA for course team accounts

zainab-amir:
This ADR describes the decision and process of adding a new security step for course team accounts. One decision is highlighted within the doc that needs review. Any possible addition or suggestion is welcome.

Ticket: https://2u-internal.atlassian.net/browse/VAN-942

### DIFF
--- a/docs/decisions/0003-2FA-for-course-staff-accounts.rst
+++ b/docs/decisions/0003-2FA-for-course-staff-accounts.rst
@@ -1,0 +1,65 @@
+3. 2FA for Course Team Accounts
+-----------------------------------------
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Course team accounts have a great deal of access to and control over sensitive information as well as the ability to run arbitrary Javascript on the LMS. An attacker who compromises an educator account could do a large amount of damage. To protect our learners (and our reputation) we should require 2FA or other enhanced account security for all accounts with these elevated permissions.
+
+An attacker should find it difficult to gain access to the privileged aspects of an educator's account (even if in possession of their password).
+
+
+Proposed Solutions
+------------------
+
+Option 1: Enable Two Factor Authentication (2FA)
+
+Option 2: Ask users for their password before specific restricted actions on site
+
+Option 3: Reset password after a specific time period
+
+
+Decision
+--------
+
+We have decided to go with Option 1. We will enable **Two Factor Authentication** (2FA) for user accounts that belong to course teams (in some cases these can be staff accounts as well).
+
+This change can be accommodated within our current login flow. A high level flow of this process is as follow:
+
+- A user belonging to a course team will authenticate themselves like they currently do, either by providing login credentials or by SSO.
+- Once they have authenticated successfully they will be sent an automatically generated authentication code to their primary email address.
+- The user will prompted to enter the code sent to their email address.
+- After successful verification of code, assign a session to user.
+
+Why choose 2FA?
+***************
+
+- It is a conventional way to secure user accounts.
+- We can protect staff users at the first step they take on our site i.e authentication.
+- Development wise it will not require extensive effort as compared to the second option.
+
+
+Enhancements to above approach
+******************************
+
+- We could reduce session expiration time only for course team users. Currently session expiration time is four weeks and we can reduce this time to two weeks for educator accounts.
+- We can also go with the first (2FA) and third (Reset password after a specific time period) option simultaneously.
+
+Consequences
+------------
+
+Adds an additional step for some users which can negatively affect the user experience.
+
+References
+----------
+
+Corresponding Security Ticket:
+https://2u-internal.atlassian.net/browse/SEG-96
+
+Detailed analysis of all security enhancements options provided above can be found here:
+https://openedx.atlassian.net/l/c/3HKHjZY4


docs/decisions/0003-2FA-for-course-staff-accounts.rst
```diff
@@ -0,0 +1,65 @@
...
+- It is a conventional way to secure user accounts.
+- We can protect staff users at the first step they take on our site i.e authentication.
+- Development wise it will not require extensive effort as compared to the second option.
+
+
+Enhancements to above approach
+******************************
+
+- We could reduce session expiration time only for course team users. Currently session expiration time is four weeks and we can reduce this time to two weeks for educator accounts.
+- We can also go with the first (2FA) and third (Reset password after a specific time period) option simultaneously.
```
timmc-edx, line 51: I believe that the current best practice is to not reset passwords on a cadence, since that makes people more likely to pick simple passwords. (They _should_ be using a password manager, but we can't force them to.)

docs/decisions/0003-2FA-for-course-staff-accounts.rst
```diff
@@ -0,0 +1,65 @@
...
+
+- It is a conventional way to secure user accounts.
+- We can protect staff users at the first step they take on our site i.e authentication.
+- Development wise it will not require extensive effort as compared to the second option.
+
+
+Enhancements to above approach
+******************************
+
+- We could reduce session expiration time only for course team users. Currently session expiration time is four weeks and we can reduce this time to two weeks for educator accounts.
```
idegtiarov, line 50: Two weeks doesn't make a big difference compared to four weeks if the course team user's account is compromised somehow. Should it be daily or at least has a configurable period for session expiration time?

docs/decisions/0003-2FA-for-course-staff-accounts.rst
zainab-amir: We enabled HIBP compliance on password reset and registration pages hence they cannot set weak passwords. Do you still think we should not ask them to reset password?

docs/decisions/0003-2FA-for-course-staff-accounts.rst
zainab-amir: - We can make it configurable. 
- The only issue I see with making the session expiry time to 1 day would be that course team are normally making changes that might span over a few days and they might lose those edit if they don't save them as draft.

docs/decisions/0003-2FA-for-course-staff-accounts.rst
timmc-edx: Strictly speaking, we can't keep people from setting weak passwords -- we can only prevent them from using specific known-to-be-weak passwords. If `superman1` is disallowed, they might choose `superm@n1` instead, which isn't much of an improvement.

HIBP will definitely help, but I do think that we should avoid doing password resets except in cases of known compromise.

docs/decisions/0003-2FA-for-course-staff-accounts.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Decision
+--------
+
+We have decided to go with Option 1. We will enable **Two Factor Authentication** (2FA) for user accounts that belong to course teams (in some cases these can be staff accounts as well).
+
+This change can be accommodated within our current login flow. A high level flow of this process is as follow:
+
+- A user belonging to a course team will authenticate themselves like they currently do, either by providing login credentials or by SSO.
+- Once they have authenticated successfully they will be sent an automatically generated authentication code to their primary email address.
```
timmc-edx, line 35: Have you looked into other kinds of 2FA?

Email-based 2FA protects users in the situation where their Open edX password is compromised but their email account remains secure. This will be true for some users, but for others the passwords will unfortunately be the same in the two systems. Password re-use is tremendously common. For those people, email isn't truly a second factor, since they use the same password for two services (from the same device, even).

I think you can look at 2FA a few different ways, although some of these overlap:

- Whether it protects against a password compromise due to A) reuse/just being weak, B) phishing (can get one TOTP token), or C) on-device malware (can capture full TOTP secret, if stored on same device)
- Whether the password is reused for the second factor (e.g. securing the email account too)
- Whether the password is entered on the same device where the second factor is secured (malware or a thief gets both in one go)

From that perspective, here's how I see some of the options:

- **Email token**: Vulnerable if password is reused for email service, or if password was compromised via on-device malware, or in phishing. edX password is already equivalent in security to email account access due to reset-password functionality, so this is marginal as 2FA goes. Works OK if attack is based on password reuse *but* email account is secure.
- **SMS token**: Vulnerable to targeted attacks, either social engineering (number porting) or technical (cloning), and to phishing. Probably OK against non-targeted attacks. Requires a cell phone.
- **TOTP**: Vulnerable if login to edX also happens on the phone where the TOTP app is installed and malware is the attack method. (Unlikely for primarily-Studio users, much more likely for other sites.) Also vulnerable to phishing. Pretty secure otherwise, and doesn't require a phone—Yubikey can be used for TOTP, and in a pinch a desktop client can be used if the TOTP code is displayed in text under the QR code to allow manual enrollment (but this reduces the security in the malware case.)
- **U2F**: Vulnerable only to physical theft of hardware token—very secure. Even prevents phishing. But most people don't have these.

I think the email token is a good first step, but I'd like to see us support TOTP and U2F as the "preferred" methods (plus a "backup codes" feature), with SMS and email only provided as fallbacks when a user can't use the preferred methods. I suspect there's already a library that can do a good amount of this for us. Maybe I'm wrong, though.

zainab-amir: @timmc-edx thank you for these suggestions. We are going to do more discovery and update ADR accordingly.

JAAkana: Please keep in mind the support needs of different 2FA methods during discovery. I could imagine course staff needing help with installing 3rd party apps, with account recovery, with a change in phone access or email address, with university spam filters. Some of these are familiar support flows and some may need new tools/workflows.

docs/decisions/0003-2FA-for-course-staff-accounts.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+Decision
+--------
+
+We have decided to go with Option 1. We will enable **Two Factor Authentication** (2FA) for user accounts that belong to course teams (in some cases these can be staff accounts as well).
+
+This change can be accommodated within our current login flow. A high level flow of this process is as follow:
+
+- A user belonging to a course team will authenticate themselves like they currently do, either by providing login credentials or by SSO.
```
JAAkana, line 34: Is there a definition for the roles that would be appropriate for 2FA, and would that affect the best kind of 2FA for the platform?

As a user/support person, off the top of my head: 
 - course staff, administrator, and course data researcher roles have high levels of access to sensitive data/tools
 - course discussion roles are orthogonal to the above, and give some access to discussion data only
 - beta testers do not have access to learner data, just early access to content
 - a couple roles may exist that are affiliated with an organization rather than a course ("universal course researcher")

syedsajjadkazmii: Currently, we are doing 2FA for course teams and I don't think it would have any effect on the best Kinds of 2FA for the platform, we have just updated the document, please have a look. Thankyou

docs/decisions/0003-2FA-for-course-staff-accounts.rst
```diff
@@ -0,0 +1,155 @@
...
+       - Requires signal and connection
+       - Cannot use if phone is dead
+       - Requires custom integrations from providers.
+   * - U2F
+     - Using small USB, NFC or Bluetooth Low Energy (BTLE) devices often called “security keys.” To set it up on a site, you register your U2F device. And then tap on the device when logging in.
+     - - Secured
+       - Can use same U2F device for multiple sites
+     - - Needs a device
+       - Cost
+       - Support (only chrome supports U2F currently)
```
timmc-edx, line 101: Support is broader than that -- Firefox supports U2F as well. (But agreed that it can't be the only supported method, so probably not something we want to start with.)

PR #581, opened 2022-06-02
feat: [VAN-953] update progressive profiling according to MFE API

attiyaIshaque:
## Description 
Remove  ***Optional Field*** API call from the progressive profiling page. Use a `common component` to get the call. from api.

## Supporting information

Ticket: https://2u-internal.atlassian.net/browse/VAN-953

### DIFF
--- a/src/common-components/RedirectLogistration.jsx
+++ b/src/common-components/RedirectLogistration.jsx
@@ -10,7 +10,7 @@ import { setCookie } from '../data/utils';
 
 function RedirectLogistration(props) {
   const {
-    finishAuthUrl, redirectUrl, redirectToWelcomePage, success,
+    finishAuthUrl, redirectUrl, redirectToWelcomePage, success, optionalFields,
   } = props;
   let finalRedirectUrl = '';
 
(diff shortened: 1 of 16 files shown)


src/welcome/ProgressiveProfiling.jsx
```diff
@@ -20,27 +20,27 @@ import {
...
 import BaseComponent from '../base-component';
 
 const ProgressiveProfiling = (props) => {
   const {
-    extendedProfile, fieldDescriptions, formRenderState, intl, submitState, showError,
+    formRenderState, intl, submitState, showError,
   } = props;
-
+  const fieldDescriptions = props.location.state.fieldDescriptions.fields;
+  const extendedProfile = props.location.state.fieldDescriptions.extended_profile;
```
mubbsharanwar, line 43: `const fieldDescriptions = props.location.state.fieldDescriptions.fields;`
`const extendedProfile = props.location.state.fieldDescriptions.extended_profile;`
We can move this to `mapStateToProps` by using `fieldDescriptionSelector` like

`const mapStateToProps = state => ({
 fieldDescriptions: fieldDescriptionSelector(state),
 )};`
 and then get from props.

mubbsharanwar (changes requested):
Some minor changes required otherwise looks good.

src/common-components/RedirectLogistration.jsx
```diff
@@ -43,13 +52,15 @@ RedirectLogistration.defaultProps = {
   success: false,
   redirectUrl: '',
   redirectToWelcomePage: false,
+  optionalFields: null,
```
mubbsharanwar, line 55: nit:
`optionalFields: null,` to `optionalFields: {},`

mubbsharanwar (approved):
Just some nits otherwise looks good.

PR #576, opened 2022-05-18
fix: Persist Register/Sign in form states on tab switch [VAN-618]

syedsajjadkazmii:
**Before:**

When the user switched tab between the Register page, Login page, and ForgotPassword page, the entered data in the forms was lost

**After:**

Data entered in any of the forms is persisted on the switching tab between the Register page, Login page, and ForgotPassword page.

[VAN-618](https://2u-internal.atlassian.net/browse/VAN-618)

**Demonstration**

https://user-images.githubusercontent.com/52817156/170024175-ce8bc1ab-5eba-46c7-9f90-208b20c2a0bb.mov

PR #582, opened 2022-06-03
Revert "fix: Persist Register/Sign in form states on tab switch [VAN-618]"

waheedahmed:
Reverts openedx/frontend-app-authn#576


rgraber (approved):
LGTM

PR #585, opened 2022-06-06
fix: Persist Register/Sign in form states on tab switch [VAN-618]

syedsajjadkazmii:
**Before:**  

When the user switched tab from register to sign or vice versa, the entered data in the form was lost

**After:**  

Data entered in the form is persisted on tab switching.


mubbsharanwar (approved):

shafqatfarhan (approved):
Looks good 👍

PR #597, opened 2022-06-27
fix: password tooltip should use extra small font size

syedsajjadkazmii:
Password tooltip should use an extra small font size of 12px.

[VAN-577](https://2u-internal.atlassian.net/browse/VAN-577)

**Before**

Screen Shot 2022-06-27 at 4 14 34 PM

**After**

Screen Shot 2022-06-27 at 4 15 46 PM

### DIFF
--- a/src/_style.scss
+++ b/src/_style.scss
@@ -296,8 +296,10 @@ select.form-control {
 .tooltip-shadow {
   box-shadow: 0px 5px 15px 0px rgba(0,0,0,0.3) !important;
 }
+
 #password-requirement-left {
   opacity: 1;
+  @extend .x-small;
   .tooltip-inner {
     @extend .tooltip-shadow;
     background: white;
@@ -315,6 +317,7 @@ select.form-control {
   opacity: 1;
   width: 90%;
   margin-bottom: 10px;
+  @extend .x-small;
   .tooltip-inner {
     max-width: inherit;
     background: white;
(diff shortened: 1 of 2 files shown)


src/_style.scss
```diff
@@ -296,8 +296,15 @@ select.form-control {
 .tooltip-shadow {
   box-shadow: 0px 5px 15px 0px rgba(0,0,0,0.3) !important;
 }
+
+.password-tooltip {
+  font-size: 12px;
+  line-height: 2;
+}
+
```
zainab-amir, line 304: Can't we use the `.x-small` class directly on the tooltip? 
https://paragon-openedx.netlify.app/foundations/typography

zainab-amir (changes requested):

src/_style.scss
```diff
@@ -296,8 +296,15 @@ select.form-control {
 .tooltip-shadow {
   box-shadow: 0px 5px 15px 0px rgba(0,0,0,0.3) !important;
 }
+
+.password-tooltip {
+  line-height: 2;
```
zainab-amir, line 301: why are we updating the line height? Just use x-small class and it will adjust automatically to 1.25rem.

Screenshot 2022-07-13 at 4 38 50 PM

src/_style.scss
```diff
@@ -296,8 +296,15 @@ select.form-control {
...
+  @extend .x-small
+}
+
 #password-requirement-left {
   opacity: 1;
+  @extend .password-tooltip;
```
zainab-amir, line 307: no need for additional class just do this
```suggestion
  @extend .x-small;
```

zainab-amir (approved):

PR #601, opened 2022-07-19
feat: Remove discount banner code

zainab-amir:
#### Context:
Removed stale code from logistration. This includes removing:
- Top discount banner
- Side discount banner
- Toast that used to appear when user successfully copied the code
- Discount code message that was added to Registration page
- clipboard package as its not needed anymore

Ticket: https://2u-internal.atlassian.net/browse/VAN-1005


src/register/RegistrationPage.jsx
```diff
@@ -85,7 +85,7 @@ class RegistrationPage extends React.Component {
       failureCount: 0,
       startTime: Date.now(),
       totalRegistrationTime: 0,
-      optimizelyExperimentName: '',
+      optimizelyExperimentName: '', // eslint-disable-line react/no-unused-state
```
waheedahmed, line 88: Can we remove this state as well alongside `getExperiments()`? I don't think we have any new experiments to launch now.

waheedahmed (approved):
One minor comment otherwise looks good 👍

src/register/RegistrationPage.jsx
zainab-amir: okay

PR #606, opened 2022-08-04
fix: Small changes in extended profile

attiyaIshaque:
### Description
Update the error message both backend and frontend to make it consistent and also makes some fixes.

[VAN-976](https://2u-internal.atlassian.net/browse/VAN-976)

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -69,7 +69,6 @@ class RegistrationPage extends React.Component {
     // permanent part of Authn and remove extra code
     this.showDynamicRegistrationFields = getConfig().ENABLE_DYNAMIC_REGISTRATION_FIELDS;
     this.tpaHint = getTpaHint();
-    this.isRegistered = true;
     const { registrationFormData } = this.props;
     this.state = {
       country: '',
@@ -115,7 +114,7 @@ class RegistrationPage extends React.Component {
     if (this.tpaHint) {
       payload.tpa_hint = this.tpaHint;
     }
-    payload.is_registered = this.isRegistered;
+    payload.is_register_page = true;
     this.props.resetRegistrationForm();
     this.props.getThirdPartyAuthContext(payload);
   }
--- a/src/welcome/ProgressiveProfiling.jsx
+++ b/src/welcome/ProgressiveProfiling.jsx
@@ -73,12 +73,14 @@ const ProgressiveProfiling = (props) => {
     e.preventDefault();
     const authenticatedUser = getAuthenticatedUser();
     const payload = { ...values, extendedProfile: [] };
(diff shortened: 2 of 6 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -115,7 +115,7 @@ class RegistrationPage extends React.Component {
     if (this.tpaHint) {
       payload.tpa_hint = this.tpaHint;
     }
-    payload.is_registered = this.isRegistered;
+    payload.is_register_page = this.isRegistered;
```
waheedahmed, line 118: Do we need this `this.isRegistered` anywhere else? If not can we remove this and simply assign the `True` here?

src/welcome/ProgressiveProfiling.jsx
```diff
@@ -73,12 +73,14 @@ const ProgressiveProfiling = (props) => {
     e.preventDefault();
     const authenticatedUser = getAuthenticatedUser();
     const payload = { ...values, extendedProfile: [] };
-    extendedProfile.forEach(fieldName => {
-      if (values[fieldName]) {
-        payload.extendedProfile.push({ fieldName, fieldValue: values[fieldName] });
-      }
-      delete payload[fieldName];
-    });
+    if (Object.keys(extendedProfile).length !== 0) {
```
waheedahmed, line 76: Shouldn’t the loop don't execute if the object is empty? Why this check is necessary here?

src/register/RegistrationPage.jsx
attiyaIshaque: Updated.

src/welcome/ProgressiveProfiling.jsx
attiyaIshaque: when object is empty its shows error that why i have add this check.

src/register/RegistrationPage.jsx
```diff
@@ -115,7 +115,7 @@ class RegistrationPage extends React.Component {
     if (this.tpaHint) {
       payload.tpa_hint = this.tpaHint;
     }
-    payload.is_registered = this.isRegistered;
+    payload.is_register_page = true;
```
waheedahmed, line 118: Please remove `this.isRegistered's` declaration as well at L72.

src/welcome/ProgressiveProfiling.jsx
waheedahmed: I believe it can only happen if the `extendedProfile` is undefined or `NULL` but not an empty object `{}`. NVM please update the condition to `if (Object.keys(extendedProfile).length > 0)`.

attiyaIshaque:
@waheedahmed I have addressed your comment.

waheedahmed (approved):

PR #610, opened 2022-08-12
fix: Saved user credentials not filling [VAN-1027]

syedsajjadkazmii:
User credentials saved in browsers were not filled in the username field when tried to fill from the username field.

[VAN-1027](https://2u-internal.atlassian.net/browse/VAN-1027)


src/login/LoginPage.jsx
```diff
@@ -70,13 +70,16 @@ class LoginPage extends React.Component {
...
 
   shouldComponentUpdate(nextProps) {
-    if (this.props.loginFormData !== nextProps.loginFormData) {
-      if (nextProps.loginFormData) {
-        this.setState({
-          ...nextProps.loginFormData,
-        });
-        return false;
-      }
+    if (this.props.loginFormData !== nextProps.loginFormData && nextProps.loginFormData) {
```
attiyaIshaque, line 73: @syedsajjadkazmii Why we are adding same condition two times ?
`nextProps.loginFormData && nextProps.loginFormData)`

src/register/RegistrationPage.jsx
```diff
@@ -120,15 +120,17 @@ class RegistrationPage extends React.Component {
...
-    if (this.props.registrationFormData !== nextProps.registrationFormData) {
-      if (nextProps.registrationFormData) {
-        // do not set focused field's value from redux store to retain entered data in focused field
-        const { focusedField } = this.state;
-        const { [focusedField]: _, ...registrationData } = nextProps.registrationFormData;
-        this.setState({
-          ...registrationData,
-        });
-      }
+    if (this.props.registrationFormData !== nextProps.registrationFormData && nextProps.registrationFormData) {
```
attiyaIshaque, line 123: Same for here `nextProps.registrationFormData && nextProps.registrationFormData`

src/login/LoginPage.jsx
```diff
@@ -70,13 +70,16 @@ class LoginPage extends React.Component {
...
+      const credentials = {
```
shafqatfarhan, line 74:
```suggestion
      const nextState = {
```

src/register/RegistrationPage.jsx
```diff
@@ -120,15 +120,17 @@ class RegistrationPage extends React.Component {
...
+      // do not set focused field's value from redux store to retain entered data in focused field
+      const credentials = {
```
shafqatfarhan, line 125:
```suggestion
      const nextState = {
```

src/register/RegistrationPage.jsx
```diff
@@ -120,15 +120,17 @@ class RegistrationPage extends React.Component {
...
-      if (nextProps.registrationFormData) {
-        // do not set focused field's value from redux store to retain entered data in focused field
-        const { focusedField } = this.state;
-        const { [focusedField]: _, ...registrationData } = nextProps.registrationFormData;
-        this.setState({
-          ...registrationData,
-        });
-      }
+    if (this.props.registrationFormData !== nextProps.registrationFormData && nextProps.registrationFormData) {
+      // do not set focused field's value from redux store to retain entered data in focused field
```
shafqatfarhan, line 124: This comment is for focusedField logic and should be adjacent to `const { focusedField } = this.state;`

shafqatfarhan (approved):
Just a couple of suggestions, otherwise LGTM.

src/login/LoginPage.jsx
syedsajjadkazmii: @attiyaIshaque actually `this.props.loginFormData !== nextProps.loginFormData` will be evaluated first and then the result will be evaluated with ` && nextProps.loginFormData`. This is because of the operator precedence, Relational operators are evaluated before logical operators, as `!==` is the relational operator and `&&` is logical operator. 
Moreover, first condition is checking for the state updates and the second condition is just validating if `nextProps.loginFormData` is not null.

syedsajjadkazmii: Plus, I felt like there is no need for a second condition, so I removed that

attiyaIshaque (approved):

PR #616, opened 2022-08-22
fix: sign-in header on login page

mubbsharanwar:
- sign-in header on login page only show when social authn is active.

VAN-1055

### DIFF
--- a/src/login/LoginPage.jsx
+++ b/src/login/LoginPage.jsx
@@ -161,14 +161,14 @@ class LoginPage extends React.Component {
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


syedsajjadkazmii (approved):

waheedahmed (commented):
Wasn't this to hide everything inside the other sign-in options if there is any active pipeline alongside the header text?

mubbsharanwar:
> Wasn't this to hide everything inside the other sign-in options if there is any active pipeline alongside the header text?

We update the condition. Now if other pipelines will be active this dos't hide those.

PR #621, opened 2022-08-30
feat: VAN-1056 - remove formik

zainab-amir:
- Remove Formik
- Updated ForgotPasswordAlert to use constants
- Fixed design issues

Ticket: https://2u-internal.atlassian.net/browse/VAN-1056

### DIFF
--- a/src/forgot-password/ForgotPasswordPage.jsx
+++ b/src/forgot-password/ForgotPasswordPage.jsx
@@ -13,7 +13,6 @@ import {
   Tabs,
 } from '@edx/paragon';
 import { ChevronLeft } from '@edx/paragon/icons';
-import { Formik } from 'formik';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { Redirect } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


src/forgot-password/ForgotPasswordPage.jsx
```diff
@@ -28,16 +27,15 @@ import ForgotPasswordAlert from './ForgotPasswordAlert';
...
   } = props;
 
-  const platformName = getConfig().SITE_NAME;
-  const supportUrl = getConfig().LOGIN_ISSUE_SUPPORT_LINK;
-  const regex = new RegExp(VALID_EMAIL_REGEX, 'i');
-
-  const [bannerEmail, setBannerEmail] = useState('');
   const [email, setEmail] = useState(props.email);
+  const [bannerEmail, setBannerEmail] = useState('');
+  const [formErrors, setFormErrors] = useState('');
```
syedsajjadkazmii, line 38: Can we use `validationError` state instead of introducing `formErrors` because we can achieve our goal for `ForgetPasswordAlert` with it too?

zainab-amir: No. If we use validationError it will continuously show the error alert but we only want to show it once the user has submitted the form.
Also if the validation error changes, the header error will also change which should not happen.

syedsajjadkazmii: Gotcha. Thankyou!

syedsajjadkazmii (approved):
LGTM 👍🏻

PR #642, opened 2022-09-12
fix: username prefixed with space character

zainab-amir:
|Before|After|
|----|-------|
Screen Shot 2022-09-12 at 4 55 17 PM|Screen Shot 2022-09-12 at 4 54 34 PM|

Ticket: https://2u-internal.atlassian.net/browse/VAN-1087

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -317,11 +317,19 @@ class RegistrationPage extends React.Component {
   }
 
   handleOnChange = (e) => {
-    if (!(e.target.name === 'username' && e.target.value.length > 30)) {
-      this.setState({
-        [e.target.name]: e.target.value,
-      });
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -317,11 +317,22 @@ class RegistrationPage extends React.Component {
...
+        this.setState({
+          [name]: value.substring(1),
+        });
+        return;
+      }
     }
+
+    this.setState({
+      [name]: value,
+    });
```
syedsajjadkazmii, line 335: Looks good to me, just a suggestion if we can have setState in one place only? like this
```suggestion
      if (value.startsWith(' ')) {
          value = value.substring(1),
      }
    }

    this.setState({
      [name]: value,
    });
```

syedsajjadkazmii (approved):

PR #648, opened 2022-09-23
feat: Add tpa auth and institution login

syedsajjadkazmii:
The PR has the following changes: 

- Add code to handle third-party auth registration and institution registration flows
- Add error handling
- Password field is not shown on the form when a user authenticates via a TPA link
- Handle state persistence when tabs are switched

### DIFF
--- a/src/common-components/Logistration.jsx
+++ b/src/common-components/Logistration.jsx
@@ -16,15 +16,21 @@ import { Redirect } from 'react-router-dom';
 
 import BaseComponent from '../base-component';
 import { LOGIN_PAGE, REGISTER_PAGE } from '../data/constants';
-import { getTpaHint, updatePathWithQueryParams } from '../data/utils';
+import { getTpaHint, getTpaProvider, updatePathWithQueryParams } from '../data/utils';
 import { LoginPage } from '../login';
 import { RegistrationPage } from '../register';
 import { backupRegistrationForm } from '../register/data/actions';
+import {
+  tpaProvidersSelector,
+} from './data/selectors';
 import messages from './messages';
 
 const Logistration = (props) => {
-  const { intl, selectedPage } = props;
-  const tpa = getTpaHint();
+  const { intl, selectedPage, tpaProviders } = props;
+  const tpaHint = getTpaHint();
+  const {
+    providers, secondaryProviders,
+  } = tpaProviders;
   const [institutionLogin, setInstitutionLogin] = useState(false);
   const [key, setKey] = useState('');
 
@@ -65,6 +71,11 @@ const Logistration = (props) => {
     </div>
   );
 
+  const isValidTpaHint = () => {
+    const { provider } = getTpaProvider(tpaHint, providers, secondaryProviders);
+    return !!provider;
+  };
+
   return (
     <BaseComponent>
       <div>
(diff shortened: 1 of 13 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -486,7 +583,27 @@ RegistrationPage.propTypes = {
...
+      name: PropTypes.string,
+      firstName: PropTypes.string,
+      lastName: PropTypes.string,
+      username: PropTypes.string,
+    }),
+  }),
   validateFromBackend: PropTypes.func.isRequired,
+  institutionLogin: PropTypes.bool.isRequired,
+  handleInstitutionLogin: PropTypes.func.isRequired,
+  userPipelineDataLoaded: PropTypes.bool,
```
zainab-amir, line 606: keep actions at the bottom and also arrange in ascending order.

src/register/RegistrationPage.jsx
```diff
@@ -510,8 +627,18 @@ RegistrationPage.defaultProps = {
   registrationResult: null,
   shouldBackupState: false,
   submitState: DEFAULT_STATE,
+  thirdPartyAuthApiStatus: 'pending',
```
zainab-amir, line 630: use constant `PENDING_STATE`

zainab-amir (commented):
I haven't tested it but these are just comments from looking at the PR. I'll test it add any more comments if needed.

src/register/RegistrationPage.jsx
```diff
@@ -36,20 +49,28 @@ const RegistrationPage = (props) => {
...
     // Actions
     backupFormState,
+    setUserPipelineDetailsLoaded,
     getRegistrationDataFromBackend,
+    userPipelineDataLoaded,
     validateFromBackend,
   } = props;
 
   const queryParams = useMemo(() => getAllPossibleQueryParams(), []);
+  const tpaHint = useMemo(() => getTpaHint(), []);
```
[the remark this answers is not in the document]
syedsajjadkazmii: I think this will work but I have some points.

- useMemo hook is used to memoize the result of an expensive computing function so that its value is returned on every render rather than again computing. since in your suggestion, we will be passing an object and the hook will memoize the object, not the functions. and I think this will be a contradiction to what the useMemo hook is used for. although it may work.

- supposing it is working great, but in our scenario, we are not passing any dependency, that means in-case if we ever needed to put dependency for a single function like for `getTpaHint()` function, then the other two functions will be running along with the desired function failing the purpose of useMemo.

- plus it's not any efficiency problem to have more than one useMemo hook, instead, I think it's more readable.

I am more inclined towards the first point where the hook will be memoizing the object instead of function. but I still need your point of view if my points are valid or not.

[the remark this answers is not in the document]
zainab-amir: Lets discuss on huddle

src/common-components/Logistration.jsx
```diff
@@ -16,15 +16,21 @@ import { Redirect } from 'react-router-dom';
...
 import messages from './messages';
 
 const Logistration = (props) => {
-  const { intl, selectedPage } = props;
-  const tpa = getTpaHint();
+  const { intl, selectedPage, tpaProvidersAndSecondaryProviders } = props;
+  const tpaHint = getTpaHint();
+  const {
+    providers, secondaryProviders,
+  } = tpaProvidersAndSecondaryProviders;
```
zainab-amir, line 33: Do we have the providers when we first load the logistration? What if the call to to MFEContext took time, will it still work?

src/common-components/Logistration.jsx
```diff
@@ -101,14 +110,29 @@ Logistration.propTypes = {
...
+    secondaryProviders: [],
+  },
 };
 
 Logistration.defaultProps = {
   selectedPage: REGISTER_PAGE,
 };
 
+const mapStateToProps = state => ({
+  tpaProvidersAndSecondaryProviders: tpaProvidersAndSecondaryProvidersSelector(state),
```
zainab-amir, line 131: Lets change the name for this. TPA stands for third party auth. We can just name is as follows:
```suggestion
  tpaProviders: tpaProvidersSelector(state),
```

src/common-components/Logistration.jsx
syedsajjadkazmii: I have provided default values for providers and secondary providers as empty list in common-components reducer, now suppose the logistration component is loaded and the MFEContext response is not received yet, in this case the `isValidTpaHint()` function will return false and the header will be displayed, and when the MFEContext response is received, the form will be rendered.

zainab-amir (approved):

PR #652, opened 2022-10-05
fix: Password reset success msg has been shown

attiyaIshaque:


syedsajjadkazmii (changes requested):
I have tested this, and the mentioned issue is resolved, thanks for resolving it but it has a ripple effect, the `Password reset success` banner persisted when switching between the sign-in and register tabs, which should not be the case. The banner should be removed once the user switches tabs.

shafqatfarhan (approved):

syedsajjadkazmii (approved):

PR #681, opened 2022-11-21
fix: use paragon branding variables for colors

shahbaz-arbisoft:
We currently have branding issue in our repo. We are hardcoding branding colors in different parts of our code.
Now, we are using paragon branding variables instead of hardcoding the colors.

Tickets: [VAN-1163](https://2u-internal.atlassian.net/browse/VAN-1163)

### DIFF
--- a/src/sass/_style.scss
+++ b/src/sass/_style.scss
@@ -82,31 +82,15 @@ $elevation-level-2-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.15);
 .alert-link {
   font-weight: normal;
   text-decoration: underline;
-  color: #0075b4 !important;
+  color: $info-300 !important;
 
   &:hover {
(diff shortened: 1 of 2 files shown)


src/sass/_style.scss
```diff
@@ -393,7 +393,7 @@ select.form-control {
     line-height: 1.25rem;
     overflow-y: scroll;
     position: absolute;
-    background-color: #fff;
+    background-color: white;
```
arbrandes, line 396: Do you mind using `$white` here and elsewhere?

arbrandes (changes requested):
This is awesome, thanks @shahbaz-arbisoft!  I only had one comment.

arbrandes (approved):
Awesome, thanks!

zainab-amir:
Update: Waiting for UX to approve the color changes.

PR #722, opened 2023-01-03
fix: uncaught exception on progressive profiling

syedsajjadkazmii:
### Description

frontend-platform's `ensureAuthenticatedUser` was throwing exceptions incase of the unauthenticated user along with redirecting to the login page. This PR passes that exception.

[VAN-1224](https://2u-internal.atlassian.net/browse/VAN-1224)

### DIFF
--- a/src/common-components/data/reducers.js
+++ b/src/common-components/data/reducers.js
@@ -23,15 +23,21 @@ const reducer = (state = defaultState, action) => {
         ...state,
         thirdPartyAuthApiStatus: PENDING_STATE,
       };
-    case THIRD_PARTY_AUTH_CONTEXT.SUCCESS:
+    case THIRD_PARTY_AUTH_CONTEXT.SUCCESS: {
+      const extendedProfile = action.payload.optionalFields.extended_profile;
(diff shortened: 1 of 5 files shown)


src/common-components/data/reducers.js
```diff
@@ -23,15 +23,21 @@ const reducer = (state = defaultState, action) => {
         ...state,
         thirdPartyAuthApiStatus: PENDING_STATE,
       };
-    case THIRD_PARTY_AUTH_CONTEXT.SUCCESS:
+    case THIRD_PARTY_AUTH_CONTEXT.SUCCESS: {
+      const extendedProfile = action.payload.optionalFields.extended_profile;
+      const extendedProfileArray = extendedProfile && JSON.parse(extendedProfile.replaceAll('\'', '"'));
```
attiyaIshaque, line 28: Looks good to me, but one question why we replacing `\` to `"`?

syedsajjadkazmii: The MFE context response we are getting is in `extended_profile="['city', 'state']"` format, and JSON parse cannot parse `'` single quotation marks, therefore it has to be converted to `"` double quotation marks before parsing.

attiyaIshaque (approved):
Please rebase with master branch before merging.

PR #751, opened 2023-02-15
Algolia configuration for personalized recommendations

mubbsharanwar:
### Description

Configured algolia for course recommendations based on user location or education level.

#### How to test locally?

- Enable progressive profiling from backend and frontend

- Enable recommendations

- Set Algolia keys

- Create new user, select education level and recommendations will be displayed 

**Note:**
Connect VPN to change IP for geolocation recommendations.


src/data/constants.js
```diff
@@ -36,3 +36,13 @@ export const INVALID_NAME_REGEX = /[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{
 // Query string parameters that can be passed to LMS to manage
 // things like auto-enrollment upon login and registration.
 export const AUTH_PARAMS = ['course_id', 'enrollment_action', 'course_mode', 'email_opt_in', 'purchase_workflow', 'next', 'save_for_later', 'register_for_free'];
+
+// Education difficulty level mapping
+export const EDUCATION_LEVEL_MAPPING = {
```
syedsajjadkazmii, line 41: could you please guide me on what is p, m, b, a, hs, jhs in it? is it the values that we get from the field? also, I am thinking is there any better way to map characters to the same level to avoid code duplication?

src/recommendations/data/reducers.js
```diff
@@ -0,0 +1,27 @@
...
+  isLoading: true,
+  showError: false,
+};
+
+const reducer = (state = defaultState, action) => {
+  switch (action.type) {
+    case USER_RECOMMENDATIONS.SUCCESS:
+      return {
+        ...state,
+        recommendations: action.payload.recommendations,
```
syedsajjadkazmii, line 14: is the recommendations coming from Algolia according to the format that we need in our components?

src/recommendations/RecommendationsPage.jsx
```diff
@@ -1,87 +1,29 @@
...
-      },
-    ],
-    title: 'Machine Learning with Python: from Linear Models to Deep Learning',
-  },
-];
-
 const RecommendationsPage = (props) => {
-  const { intl, location } = props;
+  const { intl, personalizedRecommendations, location } = props;
+  const { recommendations, isLoading, showError } = personalizedRecommendations;
```
syedsajjadkazmii, line 18: we can also combine there two destructing. Could you please use nested destructing?

src/recommendations/RecommendationsPage.jsx
```diff
@@ -1,87 +1,29 @@
...
   const DASHBOARD_URL = getConfig().LMS_BASE_URL.concat(DEFAULT_REDIRECT_URL);
+  useEffect(() => {
+    let isCalled = false;
```
syedsajjadkazmii, line 21: what is the purpose of `isCalled`?

src/recommendations/RecommendationsPage.jsx
```diff
@@ -1,87 +1,29 @@
...
+    if(isCalled && (showError || recommendations.length < 3)) {
```
syedsajjadkazmii, line 22: what if there was no error and the recommendations were less than 3?

syedsajjadkazmii (changes requested):

src/recommendations/data/reducers.js
```diff
@@ -0,0 +1,27 @@
+import { USER_RECOMMENDATIONS } from './actions';
+
+export const defaultState = {
+  recommendations: [],
+  isLoading: true,
```
syedsajjadkazmii, line 5: can we use the approach for API status that is being used in the common components reducer for thirdPartyAuthContext?

src/data/constants.js
mubbsharanwar: Yes. we get these values from education level field and we required this mapping for algolia query. We can reduce this duplication but then we need to write some extra code which will also increase  the complexity.

src/recommendations/data/reducers.js
mubbsharanwar: Yes I coordinated with Attiya and She was ok with the current algolia response

src/recommendations/RecommendationsPage.jsx
mubbsharanwar: Our component is rendering more then one time, to stop this behavior i used this. I am looking for bater solutions how we can stop multiple rendering.

src/recommendations/data/reducers.js
mubbsharanwar: Actually we can remove this state. We are async calling algolia and on progressive profiling page we are already handling loading state.
Whats your thought on it?

syedsajjadkazmii: lets call on this.

syedsajjadkazmii: how is it linked to the loading state of progressive profiling page?

src/progressive-profiling/data/sagas.js
```diff
@@ -13,9 +19,17 @@ export function* saveUserProfileInformation(action) {
...
     yield call(patchAccount, action.payload.username, action.payload.data);
 
+    const personalizedRecommendations = yield call(
+      getPersonalizedRecommendations,
+      EDUCATION_LEVEL_MAPPING[action.payload.data.level_of_education],
+    );
+
     yield put(saveUserProfileSuccess());
+
+    yield put(getUserRecommendationsSuccess(personalizedRecommendations));
```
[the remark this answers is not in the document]
mubbsharanwar: not needed now due to approach change.

src/recommendations/data/service.js
```diff
@@ -0,0 +1,22 @@
...
+
+  const { hits } = await index.search('', {
+    aroundLatLngViaIP: true,
+    facetFilters: [
+      'product:Course', 'availability:Available now', educationLevel,
+    ],
+  });
+
+  return camelCaseObject(hits);
+};
```
zainab-amir, line 20: If education level is not present we should not send it.  Also lets move the filter into a variable so that its easier to manipulate
```suggestion
const getPersonalizedRecommendations = async (educationLevel) => {
  const facetFilters = ['product:Course', 'availability:Available now'];
  
  if (educationLevel) {
    facetFilters.push(`level:${educationLevel}`);
  }

  const client = algoliasearch(process.env.ALGOLIA_APP_ID, process.env.ALGOLIA_SEARCH_KEY);
  const index = client.initIndex(INDEX_NAME);

  const { hits } = await index.search('', {
    aroundLatLngViaIP: true,
    facetFilters,
  });

  return camelCaseObject(hits);
};
```

PR #754, opened 2023-02-21
feat: add optimizely exp for recommendations page

syedsajjadkazmii:
VAN-1294

### Description

This PR adds optimizely experiment to recommendations page. 
[VAN-1294](https://2u-internal.atlassian.net/browse/VAN-1294)

Tested locally with optimizely keys


src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -74,6 +75,18 @@ const ProgressiveProfiling = (props) => {
...
+  useEffect(() => {
+    let timer = null;
+    if (showRecommendationsPage) {
+      activateRecommendationsExperiment();
+      timer = setTimeout(() => {
+        setShowRecommendationsPage(isUserInVariation());
+      }, 500);
+    }
+    return () => clearTimeout(timer);
+  }, [showRecommendationsPage]);
```
zainab-amir, line 88: If the recommendations are already set to true and then we activate the experiment and find that the recommendations are not enabled for this user we will be switching between "Next"/"Submit" text. 

Should we not combine the conditions? 
Maybe move the complete recommendations logic in a separate useEffect?

syedsajjadkazmii: good catch. thanks

src/recommendations/optimizelyExperiment.js
```diff
@@ -0,0 +1,21 @@
+const RECOMMENDATIONS_EXP_ID = 23367660127;
```
zainab-amir, line 1: Lets move this to settings

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -45,7 +47,7 @@ const ProgressiveProfiling = (props) => {
   const [values, setValues] = useState({});
   const [openDialog, setOpenDialog] = useState(false);
   const [showRecommendationsPage, setShowRecommendationsPage] = useState(false);
-
+  const authenticatedUser = getAuthenticatedUser();
```
zainab-amir, line 50: Will this cause any error if the user doesn't exit i.e some unauthenticated user tries to open the `/welcome` page

syedsajjadkazmii: nope. it will just return `null` to authenticated user and the `ensure_authenticated_user` function will redirect the user to login page

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -58,21 +60,30 @@ const ProgressiveProfiling = (props) => {
...
 
-    if (enablePersonalizedRecommendations && !userEnrollmentAction) {
-      setShowRecommendationsPage(true);
+  useEffect(() => {
+    let queryParams = {};
+    let timer = null;
+    if (registrationResponse) {
+      queryParams = getAllPossibleQueryParams(registrationResponse.redirectUrl);
+    }
+    if (enablePersonalizedRecommendations && !('enrollment_action' in queryParams)) {
```
zainab-amir, line 75: This will activate the experiment even if registration response is not present and the condition here will redirect the user to dashboard: https://github.com/openedx/frontend-app-authn/blob/d5513bddecb7a99dc5b9f077091b76a7f32ee676/src/progressive-profiling/ProgressiveProfiling.jsx#L88-L91

src/recommendations/RecommendationCard.jsx
```diff
@@ -20,9 +22,24 @@ const RecommendationCard = (props) => {
     return keys.slice(0, -2);
   };
 
+  const handleCardClick = () => {
+    trackRecommendationsClicked(
+      recommendation.activeRunKey,
```
zainab-amir, line 27: Can we convert course run key to course key?

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -58,21 +60,30 @@ const ProgressiveProfiling = (props) => {
...
-      }
     }
+  }, [DASHBOARD_URL, registrationResponse]);
 
-    if (enablePersonalizedRecommendations && !userEnrollmentAction) {
-      setShowRecommendationsPage(true);
+  useEffect(() => {
+    let queryParams = {};
+    let timer = null;
+    if (registrationResponse) {
```
zainab-amir, line 72:
```suggestion
    if (registrationResponse) {
      let queryParams = {};
      let timer = null;
```

src/progressive-profiling/ProgressiveProfiling.jsx
zainab-amir, line 50: Should we memoize this value?

src/progressive-profiling/ProgressiveProfiling.jsx
syedsajjadkazmii: we cant do that for timer, we need it outside the if block end, therefore I kept both timer and queryParams together.  should I move queryParams inside if?

src/recommendations/RecommendationCard.jsx
syedsajjadkazmii: we are not getting course keys from Agolia.

zainab-amir: Yes, I know. That is why I asked to convert it. We already do this here: https://github.com/openedx/edx-platform/blob/d694b70439b9dcbb724d5954a5f7a88d05eaadd3/lms/djangoapps/learner_recommendations/views.py#L112-L113
Maybe see if prospectus already has a way of doing it otherwise you can use regex.

src/progressive-profiling/ProgressiveProfiling.jsx
zainab-amir: I guess than outside is fine

src/recommendations/RecommendationCard.jsx
syedsajjadkazmii: yes we can do that

src/progressive-profiling/ProgressiveProfiling.jsx
syedsajjadkazmii: I don't think we should memoize it since the get authenticated user JWT tokens refresh token for authentication and we do want to monitor live authentication for the cases where the user gets unauthenticated for some reason. right?
memoizing it will return the fetched value unless the whole component is not re-mounted or we provide a specific dependency. Also, the purpose of memoizing is the memoize the complex time taking calculations, not the API calls. what do you say?

src/recommendations/RecommendationsList.jsx
```diff
@@ -41,10 +44,12 @@ RecommendationsList.propTypes = {
     })),
     marketingUrl: PropTypes.string.isRequired,
   })),
+  userId: PropTypes.number,
 };
 
 RecommendationsList.defaultProps = {
   recommendations: [],
+  userId: null,
```
mubbsharanwar, line 52: should default value not be number ?

src/recommendations/RecommendationCard.jsx
```diff
@@ -20,9 +22,24 @@ const RecommendationCard = (props) => {
     return keys.slice(0, -2);
   };
 
+  const handleCardClick = () => {
+    trackRecommendationsClicked(
```
mubbsharanwar, line 26: Position of argus is not in correct order.

mubbsharanwar (approved):
Posted some suggestions, otherwise looks good.

src/recommendations/RecommendationsList.jsx
zainab-amir: No not a number because number can be the actual user id.

src/recommendations/RecommendationCard.jsx
syedsajjadkazmii: good catch. thanks @mubbsharanwar

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -58,21 +60,30 @@ const ProgressiveProfiling = (props) => {
...
+      queryParams = getAllPossibleQueryParams(registrationResponse.redirectUrl);
+      if (enablePersonalizedRecommendations && !('enrollment_action' in queryParams)) {
```
zainab-amir, line 74: We need to add cookie logic for testing on stage.

zainab-amir (approved):

src/progressive-profiling/ProgressiveProfiling.jsx
syedsajjadkazmii: we will testing on stage without cookie logic.

PR #755, opened 2023-02-21
zamir/van 1295/add tests for recommender

zainab-amir:
### Description

Add tests for the recommendations functionality. Tests added for
- Progressive profiling page: This is where we decide whether to add recommendations or not
- Recommendations page tests
- 

#### Merge Checklist

### DIFF
--- a/src/recommendations/RecommendationsPage.jsx
+++ b/src/recommendations/RecommendationsPage.jsx
@@ -7,31 +7,34 @@ import {
 } from '@edx/paragon';
 import PropTypes from 'prop-types';
 
-import { DEFAULT_REDIRECT_URL, EDUCATION_LEVEL_MAPPING } from '../data/constants';
+import { DEFAULT_REDIRECT_URL } from '../data/constants';
+import { EDUCATION_LEVEL_MAPPING, RECOMMENDATIONS_COUNT } from './data/constants';
 import getPersonalizedRecommendations from './data/service';
 import messages from './messages';
 import RecommendationsList from './RecommendationsList';
 
 const RecommendationsPage = (props) => {
   const { intl, location } = props;
+  const registrationResponse = location.state?.registrationResult;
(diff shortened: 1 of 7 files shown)


src/recommendations/RecommendationsPage.jsx
```diff
@@ -47,7 +50,7 @@ const RecommendationsPage = (props) => {
     }
   };
 
-  if (!isLoading && recommendations.length < 3) {
+  if (!isLoading && recommendations.length < RECOMMENDATIONS_COUNT) {
```
syedsajjadkazmii, line 53: Don't we want to show 3 recommendations?

src/recommendations/tests/RecommendationsPage.test.jsx
```diff
@@ -0,0 +1,180 @@
...
+const IntlRecommendationsPage = injectIntl(RecommendationsPage);
+const mockStore = configureStore();
+
+jest.mock('../data/service');
+
+describe('RecommendationsPageTests', () => {
+  let defaultProps = {};
+  let store = {};
+
+  const mockedResponse = [
```
syedsajjadkazmii, line 23: can you put this mocked constant data into a separate file such as `tests/constants.js`. This approach is used by learner home and I think its a good approach to separate constants from tests.

src/recommendations/RecommendationsPage.jsx
zainab-amir: I am not sure where this `3` came from so for safe side I have used `RECOMMENDATIONS_COUNT` and have also created a thread with UX for confirmation. They are finalizing a strategy where we will backfill the missing recommendations with general ones but for now lets keep it the condition as mentioned above.

syedsajjadkazmii (commented):
I don't see any test for the loader state. could you please add that too?

syedsajjadkazmii (approved):
Thanks

PR #771, opened 2023-03-08
Card banner image is not centered aligned

ahtesham-quraish:
VAN-1325

### Description

Post recommendation card's banner image was not centered aligned. Now added the object position property to make centered so that it looks like what we have in edx.org page.

Tested locally

### DIFF
--- a/src/sass/_recommendations_page.scss
+++ b/src/sass/_recommendations_page.scss
@@ -39,6 +39,7 @@
 
   .pgn__card-image-cap {
     height: 6.5rem;
+    // Property that center the image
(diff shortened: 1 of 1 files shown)


attiyaIshaque:
@ahtesham-quraish These are different types of commits. You can find more details about which to use here: https://open-edx-proposals.readthedocs.io/en/latest/oep-0051-bp-conventional-commits.html

mubbsharanwar (approved):

mubbsharanwar (approved):
nit: remove the comment and rebase before merging.

PR #773, opened 2023-03-09
fix: country dropdown issue on safari

ahtesham-quraish:
VAN-1317

### Description
On Safari browser country dropdown field is not functional so we have added the onMuseDown event for default prevention.


syedsajjadkazmii (changes requested):
Could you please add its test also?

ahtesham-quraish:
> Could you please add its test also?

We don't need any test here because its not big change code change we can skip it for now.

PR #777, opened 2023-03-13
 fix: remove validation which comes from backend after clicking regis…

ahtesham-quraish:
…tration button

VAN-1716

### Description
clear backend validation on focus of field after submission of registration of page.

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -30,6 +30,7 @@ import {
 import ConfigurableRegistrationForm from './ConfigurableRegistrationForm';
 import {
   backupRegistrationFormBegin,
+  clearRegistertionBackendError,
   clearUsernameSuggestions,
   fetchRealtimeValidations,
   registerNewUser,
@@ -58,6 +59,7 @@ const RegistrationPage = (props) => {
     intl,
     institutionLogin,
     optionalFields,
+    registrationError,
     registrationErrorCode,
     registrationResult,
     shouldBackupState,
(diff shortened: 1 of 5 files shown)


src/register/data/tests/reducers.test.js
```diff
@@ -102,6 +103,19 @@ describe('Registration Reducer Tests', () => {
...
+    const state = {
+      ...defaultState,
+      registrationError: { email: 'This email is already associated with an existing or previous Your Platform Name Here account' },
+    };
+    const action = {
+      type: CLEAR_REGISTER_BACKEND_ERROR,
+      payload: 'email',
+    };
+
+    expect(reducer(state, action)).toEqual(state);
```
mubbsharanwar, line 116: The error state should be null in the reducer while dispatching this action.

First, we need to set error state and then by dispatching our action, reducer state should be set to default.

src/register/RegistrationPage.jsx
```diff
@@ -387,6 +389,7 @@ const RegistrationPage = (props) => {
   const handleOnFocus = (event) => {
     const { name, value } = event.target;
     setErrors(prevErrors => ({ ...prevErrors, [name]: '' }));
+    clearBackendError(name);
```
mubbsharanwar, line 392: We need to fire this action on `onChange` event when our validation API return 403.
And this should be fire for all those fields for those we are enabling backend validations like email, username, and password.

ahtesham-quraish: @mubbsharanwar I have discussed this with Sajad and he told me the exact error and when it comes so instead of onchange method it should be called on onfocus. we can discuss it on call

syedsajjadkazmii (approved):

PR #779, opened 2023-03-14
feat: hide signup page on the bases of flag

attiyaIshaque:
### Description
Hides Signup page and “Login/Register” tab when ALLOW_PUBLIC_ACCOUNT_CREATION is False.

Screenshot 2023-03-17 at 2 17 26 PM

#### Merge Checklist

#### Post-merge Checklist

### DIFF
--- a/src/common-components/Logistration.jsx
+++ b/src/common-components/Logistration.jsx
@@ -33,6 +33,7 @@ const Logistration = (props) => {
   } = tpaProviders;
   const [institutionLogin, setInstitutionLogin] = useState(false);
   const [key, setKey] = useState('');
+  const disablePublicAccountCreation = getConfig().ALLOW_PUBLIC_ACCOUNT_CREATION === false;
 
   useEffect(() => {
     const authService = getAuthService();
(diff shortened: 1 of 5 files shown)


src/common-components/Logistration.jsx
```diff
@@ -79,28 +80,44 @@ const Logistration = (props) => {
...
-        </div>
+          : (
+            <div>
+              {institutionLogin
+                ? (
+                  <Tabs defaultActiveKey="" id="controlled-tab" onSelect={handleInstitutionLogin}>
+                    <Tab title={tabTitle} eventKey={selectedPage === LOGIN_PAGE ? LOGIN_PAGE : REGISTER_PAGE} />
+                  </Tabs>
+                )
+                : (!isValidTpaHint() && (
```
syedsajjadkazmii, line 98: The `isValidTpaHint ` condition is also in the else case, what if it has to do something along with `allowPublicAccountCreation = false`? I did not get a chance to test this one. but did you test that it works fine with these conditions?

syedsajjadkazmii: I think when `allowPublicAccountCreation` is true then third-party authentication should be disabled because if we try to login a user with a non-existing email through social auth such as google, then it will allow the user to create a new account, which we don't want to allow it.

src/MainApp.jsx
```diff
@@ -37,6 +38,8 @@ const MainApp = () => (
         <Redirect to={updatePathWithQueryParams(REGISTER_PAGE)} />
       </Route>
       <UnAuthOnlyRoute exact path={LOGIN_PAGE} render={() => <Logistration selectedPage={LOGIN_PAGE} />} />
+      { !allowPublicAccountCreation
```
syedsajjadkazmii, line 41: can we use the identifier `disablePublicAccountCreation` instead of `allowPublicAccountCreation` which will increase readability in my opinion. but its up to you.

src/common-components/Logistration.jsx
```diff
@@ -79,28 +80,54 @@ const Logistration = (props) => {
...
+                {institutionLogin
+                  ? (
+                    <Tabs defaultActiveKey="" id="controlled-tab" onSelect={handleInstitutionLogin}>
+                      <Tab title={tabTitle} eventKey={LOGIN_PAGE} />
+                    </Tabs>
+                  ) : (
+                    <h3 className="mb-4.5">{intl.formatMessage(messages['logistration.sign.in'])}</h3>
+                  )}
+                <LoginPage institutionLogin={institutionLogin} handleInstitutionLogin={handleInstitutionLogin} />
+              </div>
```
syedsajjadkazmii, line 97: It should be this, it was causing little UI disruption at the institution login page.

```suggestion
              <Redirect to={updatePathWithQueryParams(LOGIN_PAGE)} />
              {institutionLogin && (
                <Tabs defaultActiveKey="" id="controlled-tab" onSelect={handleInstitutionLogin}>
                  <Tab title={tabTitle} eventKey={LOGIN_PAGE} />
                </Tabs>
              )}
              <div id="main-content" className="main-content">
                {!institutionLogin && (
                  <h3 className="mb-4.5">{intl.formatMessage(messages['logistration.sign.in'])}</h3>
                )}
                <LoginPage institutionLogin={institutionLogin} handleInstitutionLogin={handleInstitutionLogin} />
              </div>
```

syedsajjadkazmii (approved):

PR #781, opened 2023-03-16
feat: upgraded to node v18, added .nvmrc and updated workflows

Mashal-m:
- Added support for node v18, updated .nvmrc and workflows.

**Ticket**
[Upgrade Node JS from 16 to 18](https://github.com/openedx/frontend-app-authn/issues/780)


abdullahwaheed (approved):
looks good to me!
@zainab-amir could you please take a look as well

package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
     "@edx/frontend-component-cookie-policy-banner": "2.2.2",
-    "@edx/frontend-platform": "3.6.1",
+    "@edx/frontend-platform": "^4.0.2",
```
zainab-amir, line 37: Why restrict it to v4 only?
```suggestion
    "@edx/frontend-platform": "4.0.2",
```

zainab-amir (commented):
Please squash your commits

package.json
Mashal-m: Because we released the node18 upgrade in frontend-platform version 4.

zainab-amir: Adding the caret (^) will update to all future minor/patch versions, without incrementing the major version. We should not restrict it unless we have reason to do it.

Mashal-m: removed

Mashal-m:
> Please squash your commits

Squashed.

PR #787, opened 2023-03-24
fix: fix recommendations viewed event count anomly

syedsajjadkazmii:
### Description

This PR fixes two anomalies with welcome page recommendations viewed event

1. Recommendations viewed event was being sent two times for original variation (one without user id and one with user id)
2. Recommendations viewed event was not having course keys for experiment variation

Tested locally

### DIFF
--- a/src/recommendations/RecommendationsPage.jsx
+++ b/src/recommendations/RecommendationsPage.jsx
@@ -24,6 +24,7 @@ const RecommendationsPage = (props) => {
 
   const [isLoading, setIsLoading] = useState(true);
   const [recommendations, setRecommendations] = useState([]);
+  const [algoliaRecommendations, setAlgoliaRecommendations] = useState([]);
   const educationLevel = EDUCATION_LEVEL_MAPPING[location.state?.educationLevel];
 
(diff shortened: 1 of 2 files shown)


src/recommendations/RecommendationsPage.jsx
```diff
@@ -24,6 +24,7 @@ const RecommendationsPage = (props) => {
 
   const [isLoading, setIsLoading] = useState(true);
   const [recommendations, setRecommendations] = useState([]);
+  const [algoliaRecommendations, setAlgoliaRecommendations] = useState([]);
```
zainab-amir, line 27: Lets use boolean value instead of saving the whole list in two places. 
```suggestion
  const [isAlgoliaRecommendation, setIsAlgoliaRecommendations] = useState(false);
```

zainab-amir: Set `isAlgoliaRecommendation` only when we get the recommendations from Algolia and then use it in the useEffect along with already existing `recommendations` state.

zainab-amir (approved):

PR #793, opened 2023-03-28
feat: track recommendations group

zainab-amir:
### Description

Added an event `edx.bi.user.recommendations.group` to track the recommendations group. This is a debugging event and will be removed later.


syedsajjadkazmii (commented):
This looks good. but as we are debugging, can we also add recommendation viewed event for optimizely? and will see if the visitors are equal to optimizely's recommendation viewed event count.

syedsajjadkazmii (approved):

PR #818, opened 2023-03-30
revert: restore the segment page event call to its previous state

shahbaz-arbisoft:
This pull request restores the previous version of the segment page event call. The previous version was changed because the page event was being called with a null user_id on the welcome page.

> Reference PR #784


shahbaz-arbisoft:
Prior to reviewing this pull request, we need to test these changes on the upcoming version of the frontend-platform. Once this testing is complete, we can decide whether or not to proceed with reviewing and merging this pull request.

PR #825, opened 2023-03-31
fix: add new flag in autoupdate action

ahtesham-quraish:
Description
Trigger autoupdate workflow when autoupdate label is added into PR

How Has This Been Tested?
This would be tested on actually PR

Related to: https://2u-internal.atlassian.net/browse/VAN-1349

### DIFF
--- a/.github/workflows/autoupdate-pull-request.yml
+++ b/.github/workflows/autoupdate-pull-request.yml
@@ -3,9 +3,17 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [ labeled ]  
+    branches:
+      - master  
 jobs:
   autoupdate:
     name: autoupdate
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.action == 'labeled' &&
+      github.event.label.name == 'autoupdate'
     runs-on: ubuntu-22.04
     steps:
       - uses: docker://chinthakagodawita/autoupdate-action:v1


.github/workflows/autoupdate.yml
```diff
@@ -3,9 +3,12 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [ labeled ]
```
shahbaz-arbisoft, line 7: It's redundant to use both `types` and `PR_FILTER` because they both serve the same purpose of filtering the pull requests that should trigger the workflow.

.github/workflows/autoupdate.yml
```diff
@@ -3,9 +3,12 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [ labeled ]    
 jobs:
   autoupdate:
     name: autoupdate
+    if: ${{ github.event.label.name == 'autoupdate' }}
```
shahbaz-arbisoft, line 11: It is redundant to have both `PR_LABELS` and the `if` condition that checks for the label name being "autoupdate".

shahbaz-arbisoft:
You can do something like this, _verify before adding_.

```
on:
  pull_request:
    branches:
      - master
```

The branches field is present in the on.pull_request field to ensure that the pull request originates from the master branch.

Or you can do this but remove the `PR_FILTER` from envs variable.
```
on:
  pull_request:
    types: [labeled]
    branches:
      - master
```

.github/workflows/autoupdate.yml
shahbaz-arbisoft, line 7: Use this.

```
pull_request:
    types: [labeled]
    branches:
      - master
```

The branches field is present in the on.pull_request field to ensure that the pull request originates from the master branch.

.github/workflows/autoupdate.yml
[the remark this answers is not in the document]
shahbaz-arbisoft: To ensure that the `autoupdate` job runs only on pull requests with the `autoupdate`

.github/workflows/autoupdate.yml
ahtesham-quraish: addressed

.github/workflows/autoupdate.yml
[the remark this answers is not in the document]
ahtesham-quraish: addressed

shahbaz-arbisoft (approved):

shahbaz-arbisoft:
Please rebase your branch before merging.

PR #828, opened 2023-04-03
fix: issue of missing user_id on welcome page events

shahbaz-arbisoft:
This pull request resolves an issue where user_ids were missing from welcome page events because of an in-page redirect. To fix this, we are now capturing user_ids after registration by making a segment identify call.

Ticket: [VAN-1262](https://2u-internal.atlassian.net/browse/VAN-1262)

### DIFF
--- a/src/progressive-profiling/ProgressiveProfiling.jsx
+++ b/src/progressive-profiling/ProgressiveProfiling.jsx
@@ -70,11 +70,11 @@ const ProgressiveProfiling = (props) => {
   }, [DASHBOARD_URL, registrationResponse]);
 
   useEffect(() => {
-    if (registrationResponse && authenticatedUser?.userId) {
+    if (ready && authenticatedUser?.userId) {
       identifyAuthenticatedUser(authenticatedUser.userId);
       sendPageEvent('login_and_registration', 'welcome');
     }
-  }, [authenticatedUser, registrationResponse]);
+  }, [authenticatedUser, ready]);
 
   useEffect(() => {
     if (registrationResponse && authenticatedUser?.userId) {


src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -59,22 +59,17 @@ const ProgressiveProfiling = (props) => {
     ensureAuthenticatedUser(DASHBOARD_URL)
       .then(() => {
         hydrateAuthenticatedUser().then(() => {
+          identifyAuthenticatedUser(authenticatedUser?.userId);
           setReady(true);
         });
       })
       .catch(() => {});
```
ahtesham-quraish, line 66: Why do we have catch block? I think we should remove it.

shahbaz-arbisoft: It was already there, I haven't added it.

ahtesham-quraish: make sense i thought you can remove it

syedsajjadkazmii: @ahtesham-quraish, if you look at the implementation of `ensureAuthenticatedUser` [here](https://github.com/openedx/frontend-platform/blob/afe3a1aa894aad413c97c315e157559d7f84be43/src/auth/AxiosJwtAuthService.js#L266-L270).

You will see that it works in a way that if the user is not authenticated, then redirect the user to the login page and throw an error along with it. This error was harmless but a useless noise in the logs as well. So we just added a catch block to don't throw errors on the console when the such scenario happens.

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -59,22 +59,17 @@ const ProgressiveProfiling = (props) => {
     ensureAuthenticatedUser(DASHBOARD_URL)
       .then(() => {
         hydrateAuthenticatedUser().then(() => {
+          identifyAuthenticatedUser(authenticatedUser?.userId);
```
syedsajjadkazmii, line 62: @shahbaz-arbisoft could you please tell why are we adding it here instead of the useEffect that we were using?

syedsajjadkazmii (approved):

PR #840, opened 2023-04-10
fix: margin issue for sso baner

ahtesham-quraish:
Description
FIX margin issue for SSO banner on login page.

How Has This Been Tested?
This is tested locally we need to login with email which is not registered before

Related to: https://2u-internal.atlassian.net/browse/VAN-1353


syedsajjadkazmii (approved):
LGTM. Thanks

PR #847, opened 2023-04-11
fix: Update eslint configuration and resolve linting errors and test …

shahbaz-arbisoft:
A recent [change](https://github.com/openedx/frontend-app-authn/pull/791) to the `eslint` version was introduced when updating the `frontend-platform` to `v12`, resulting in several linting errors. This PR aims to fix those errors. Specifically, this PR removes the `no-import-assign` rule from the [eslintrc.js](https://github.com/openedx/frontend-app-authn/blob/master/.eslintrc.js#L51) file, which was causing linting errors in the codebase.

In addition to updating the configuration, this PR also fixes various linting issues that were uncovered during testing. These fixes include updating import statements and removing unused imports. Additionally, tests were failing due to these linting issues, so this PR also addresses those test failures.

It is important to note that the `no-import-assign` rule was removed in `eslint v8.0.0`, which is the version used in this PR.

Ticket: [VAN-1357](https://2u-internal.atlassian.net/browse/VAN-1357)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,7 +48,6 @@ module.exports = createConfig('eslint', {
       },
     ],
     'function-paren-newline': 'off',
-    'no-import-assign': 'off',
     'react/no-unstable-nested-components': 'off',
   },
 });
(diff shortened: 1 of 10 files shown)


attiyaIshaque (approved):
Looks good to me 👍 Please rebase your branch with the latest code.

PR #849, opened 2023-04-11
fix: registration test coverage

ahtesham-quraish:
Description
Increase test coverage of registration flow

VAN-1368

How Has This Been Tested?
Its been test locally

Related : https://2u-internal.atlassian.net/browse/VAN-1368

### DIFF
--- a/src/register/data/tests/reducers.test.js
+++ b/src/register/data/tests/reducers.test.js
@@ -1,12 +1,13 @@
 import { getConfig } from '@edx/frontend-platform';
 
-import { DEFAULT_STATE } from '../../../data/constants';
+import { DEFAULT_REDIRECT_URL, DEFAULT_STATE, PENDING_STATE } from '../../../data/constants';
 import {
   BACKUP_REGISTRATION_DATA,
   REGISTER_CLEAR_USERNAME_SUGGESTIONS,
   REGISTER_FORM_VALIDATIONS,
   REGISTER_NEW_USER,
   REGISTER_SET_COUNTRY_CODE,
+  REGISTER_SET_USER_PIPELINE_DATA_LOADED,
   REGISTERATION_CLEAR_BACKEND_ERROR,
 } from '../actions';
 import reducer from '../reducers';
@@ -63,6 +64,42 @@ describe('Registration Reducer Tests', () => {
       },
     );
   });
+  it('should set redirect url dashboard on registration success action', () => {
+    const payload = {
+      redirectUrl: `${getConfig().BASE_URL}${DEFAULT_REDIRECT_URL}`,
+      success: true,
+    };
+    const action = {
+      type: REGISTER_NEW_USER.SUCCESS,
+      payload,
+    };
+
+    expect(reducer(defaultState, action)).toEqual(
+      {
+        ...defaultState,
+        registrationResult: payload,
+      },
+    );
+  });
+
+  it('should set the registration call and set the registration error object empty', () => {
+    const action = {
+      type: REGISTER_NEW_USER.BEGIN,
+    };
+
+    expect(reducer({
+      ...defaultState,
+      registrationError: {
+        email: 'This email already exist.',
+      },
+    }, action)).toEqual(
+      {
+        ...defaultState,
+        submitState: PENDING_STATE,
+        registrationError: {},
+      },
+    );
+  });
 
   it('should show username suggestions returned by registration error', () => {
     const payload = { usernameSuggestions: ['test12'] };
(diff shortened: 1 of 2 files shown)


src/register/data/tests/reducers.test.js
```diff
@@ -63,6 +64,37 @@ describe('Registration Reducer Tests', () => {
       },
     );
   });
+  it('on regirstration success call', () => {
+    const payload = {
+      redirectUrl: 'http://localhost:18000/dashboard',
```
mubbsharanwar, line 69:
```suggestion
      redirectUrl: `${getConfig().BASE_URL}/dashboard`,
```

src/register/data/tests/reducers.test.js
```diff
@@ -63,6 +64,37 @@ describe('Registration Reducer Tests', () => {
       },
     );
   });
+  it('on regirstration success call', () => {
```
mubbsharanwar, line 67:
```suggestion
  it('should set redirect url dashboard on registration success action', () => {
```

src/register/data/tests/reducers.test.js
```diff
@@ -63,6 +64,37 @@ describe('Registration Reducer Tests', () => {
...
+
+    expect(reducer(defaultState, action)).toEqual(
+      {
+        ...defaultState,
+        registrationResult: payload,
+      },
+    );
+  });
+
+  it('should set the registration call and set the registration error object empty', () => {
```
mubbsharanwar, line 85: Could you elaborate the purpose of this test?

src/register/data/tests/reducers.test.js
```diff
@@ -79,7 +111,68 @@ describe('Registration Reducer Tests', () => {
       },
     );
   });
+  it('should set the register user when pipline data is loaded', () => {
```
mubbsharanwar, line 114: nit: should set the registered user when SSO pipeline data is loaded

src/register/data/tests/reducers.test.js
```diff
@@ -79,7 +111,68 @@ describe('Registration Reducer Tests', () => {
...
+
+    expect(reducer(defaultState, action)).toEqual(
+      {
+        ...defaultState,
+        userPipelineDataLoaded: true,
+      },
+    );
+  });
+
+  it('dont set country code if this is already setup', () => {
```
mubbsharanwar, line 129: The test logic is setting the country but the test is don't set the country code. could you explain it?

src/register/data/tests/reducers.test.js
```diff
@@ -92,6 +185,19 @@ describe('Registration Reducer Tests', () => {
     expect(reducer(state, action)).toEqual({ ...defaultState });
   });
 
+  it('should take back data during form reset', () => {
```
mubbsharanwar, line 188: do we need this test? The state is already with default values before this action.

src/register/tests/RegistrationPage.test.jsx
```diff
@@ -384,7 +393,31 @@ describe('RegistrationPage', () => {
...
 
+      expect(registrationPage.find('input#username').prop('value')).toEqual('');
+    });
+
+    it('should give error with suggestion for common top level domain mistakes', () => {
+      const registrationPage = mount(reduxWrapper(<IntlRegistrationPage {...props} />));
+      registrationPage.find('input#email').simulate('change', { target: { value: 'ahtesham@hotmail', name: 'email' } });
+      registrationPage.find('input#email').simulate('blur');
+
+      expect(registrationPage.find('.alert-text').text()).toEqual('Did you mean ahtesham@hotmail.com?');
```
mubbsharanwar, line 408: nit: move this received message into a variable

mubbsharanwar (changes requested):

src/register/data/tests/reducers.test.js
ahtesham-quraish: addressed

src/register/data/tests/reducers.test.js
ahtesham-quraish: addressed

src/register/data/tests/reducers.test.js
ahtesham-quraish: When we initiate the call after submitting the button then this flow triggers

src/register/data/tests/reducers.test.js
ahtesham-quraish: In this we are setting the country code because that is already setup so we will skip the if condition which is written in reducer

src/register/data/tests/reducers.test.js
ahtesham-quraish: we have written because line was uncovered in reducer that is why we had to write it.

src/register/tests/RegistrationPage.test.jsx
ahtesham-quraish: You mean like this ?
const receievedMessage = 'Did you mean ahtesham@hotmail.com?';
expect(registrationPage.find('.alert-text').text()).toEqual(receievedMessage);

src/register/data/tests/reducers.test.js
mubbsharanwar: If the purpose is to increase the coverage of the action, we need to set some default value in the state then dispatch this action and verify the state should go on empty state.

mubbsharanwar (changes requested):

src/register/data/tests/reducers.test.js
mubbsharanwar: Sorry, I missed 1 thing. Could you please import dashboard from constants

src/register/data/tests/reducers.test.js
ahtesham-quraish: Addressed

src/register/data/tests/reducers.test.js
ahtesham-quraish: addressed

src/register/data/tests/reducers.test.js
ahtesham-quraish: addressed

mubbsharanwar (approved):

PR #862, opened 2023-04-17
fix: conditionally initialize optimizely

zainab-amir:
### Description

Added a check to conditionally initialize Optimizely. If the Optimizely SDK key exists, only then try to initialize it. Also updated the events and activation code to check for Optimizely instance. The code fixes following error:

Screenshot 2023-04-12 at 12 03 44 AM

#### JIRA

[VAN-1385](https://2u-internal.atlassian.net/browse/VAN-1385)

#### Merge Checklist

#### Post-merge Checklist


src/recommendations/optimizelyExperiment.js
```diff
@@ -12,17 +12,17 @@ export const eventNames = {
   * Activate the post registration recommendations optimizely experiment
   * and return the true if the user is in variation else false.
   * @param  {String} userId  user id of authenticated user.
-  * @return {string} true if the user is in variation else false
+  * @return {string} variation the user belong in
 */
-const activateRecommendationsExperiment = (userId) => optimizelyInstance.activate(RECOMMENDATIONS_EXP_KEY, userId);
+const activateRecommendationsExperiment = (userId) => optimizelyInstance?.activate(RECOMMENDATIONS_EXP_KEY, userId);
```
syedsajjadkazmii, line 17: just one question for my understanding, based on the current implementation the output of `activateRecommendationsExperiment` can be of two types now. Either it will be null or the variation name. 

Now suppose, the output was null for some reason, does that mean that the user will fall in original variation?
Would it have any effects on the experiment analytics?

zainab-amir: The current implementation is also the same. We can get three groups `null`(something went wrong with optimizely)/`original`/`treatment group`. If we get null or `original` we just don't show the recommendations.

syedsajjadkazmii (approved):

PR #865, opened 2023-04-18
fix: remove extract-react-intl-messages

ahtesham-quraish:
Remove unused package extract-react-intl-messages from package.json, this package was used to translate the messages and started using different one so we can safely remove it. We have opted to use @edx/reactifex for translations. 

PR in which this package was added. [b42ca99](https://github.com/openedx/frontend-app-authn/commit/b42ca9917680876b222990822d93d5583eb0a32e)

How Has This Been Tested?
Locally 

Related to: https://2u-internal.atlassian.net/browse/VAN-1378

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,6 @@
         "algoliasearch": "^4.14.3",
         "classnames": "2.3.2",
         "core-js": "3.30.1",
-        "extract-react-intl-messages": "4.1.1",
         "fastest-levenshtein": "1.0.16",
         "form-urlencoded": "6.1.0",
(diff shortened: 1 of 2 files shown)


syedsajjadkazmii (commented):
did you run `npm install` after this change? i believe there should be some changes to the `package-lock.json` file as well.

ahtesham-quraish:
> did you run `npm install` after this change? i believe there should be some changes to the `package-lock.json` file as well.

Yes i did @syedsajjadkazmii

syedsajjadkazmii (approved):

PR #882, opened 2023-04-26
fix: eliminate the usage of OR operator in MFE context API

shahbaz-arbisoft:
This PR removes the launch strategy logic code that was implemented to prevent the front-end from breaking while serializing the MFE context API response to camelCase on the back-end. Since both the front-end and back-end PRs have been deployed on production, the launch strategy logic code is no longer necessary.

> Related Ticket: [VAN-1311](https://2u-internal.atlassian.net/browse/VAN-1311)

### DIFF
--- a/src/common-components/data/service.js
+++ b/src/common-components/data/service.js
@@ -1,4 +1,4 @@
-import { camelCaseObject, convertKeyNames, getConfig } from '@edx/frontend-platform';
+import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
 // eslint-disable-next-line import/prefer-default-export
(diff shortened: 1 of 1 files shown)


src/common-components/data/service.js
```diff
@@ -18,10 +18,8 @@ export async function getThirdPartyAuthContext(urlParams) {
...
     });
   return {
-    fieldDescriptions: data.registrationFields || data.registration_fields || {},
-    optionalFields: data.optionalFields || data.optional_fields || {},
-    thirdPartyAuthContext: data.contextData || camelCaseObject(
-      convertKeyNames(data.context_data, { fullname: 'name' }),
-    ),
+    fieldDescriptions: data.registrationFields,
+    optionalFields: data.optionalFields,
+    thirdPartyAuthContext: data.contextData,
```
zainab-amir, line 23: Have you tested this? This will probably break the code when the MFEContext API is in the "fetching" state or has failed. We will need the default values.

zainab-amir (changes requested):

src/common-components/data/service.js
shahbaz-arbisoft: MFEContext API will have these particular keys included with default values. However, I have not yet tested the scenario where the API fails. I have just checked the test results. I will investigate this further and create a test scenario to verify the behavior in case of an API failure.

shahbaz-arbisoft: I have included a condition in the API response to handle default values.

mubbsharanwar (approved):

PR #894, opened 2023-05-08
zamir/van 1390/add basic login form

zainab-amir:
### Description

This PR includes
- Refactor of login page from class component to functional component
- Updated structure of LoginFailure component
- Account activation message banner
- Password reset success banner

#### JIRA

[VAN-1390](https://2u-internal.atlassian.net/browse/VAN-1390)

Tests will be fixed and added in a separate ticket


src/login/AccountActivationMessage.jsx
```diff
@@ -59,7 +60,7 @@ const AccountActivationMessage = (props) => {
   return activationMessage ? (
     <Alert
       id="account-activation-message"
-      className="mb-4"
+      className="mb-5"
```
syedsajjadkazmii, line 63: Just curious why would we make a style change?

syedsajjadkazmii, line 63: Just curious why would we make a style change?

src/login/data/reducers.js
```diff
@@ -1,64 +1,68 @@
...
 };
 
 const reducer = (state = defaultState, action = {}) => {
   switch (action.type) {
+    case BACKUP_LOGIN_DATA.BASE:
+      return {
+        ...state,
+        shouldBackupState: true,
+      };
+    case BACKUP_LOGIN_DATA.BEGIN:
```
syedsajjadkazmii, line 33: Should not it be `BACKUP_LOGIN_DATA.SUCCESS`?

src/login/LoginPage.jsx
```diff
@@ -1,435 +1,266 @@
...
-    return true;
-  }
+  }, [loginErrorCode]);
 
-  componentWillUnmount() {
-    if (this.props.resetPassword) {
-      this.props.loginRemovePasswordResetBanner();
+  useEffect(() => {
+    if (loginResult.success) {
+      // TODO: Do we still need this cookie?
```
syedsajjadkazmii, line 83: I don't think we are running any surveys! Maybe we should tag Wajeeha here.

src/login/LoginPage.jsx
```diff
@@ -1,435 +1,266 @@
...
-      });
-      this.props.loginRequestFailure({
-        errorCode: INVALID_FORM,
-      });
+    return { ...fieldErrors };
+  };
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    if (showResetPasswordSuccessBanner) {
```
syedsajjadkazmii, line 106: If we are only dismissing the banner here. can we do it in the `loginRequest` action inside the reducer instead of creating a whole new action?

src/login/LoginPage.jsx
```diff
@@ -1,435 +1,266 @@
...
-      [e.target.name]: e.target.value,
-    };
-    this.props.setLoginFormData(payload);
+  const handleOnFocus = (event) => {
+    const { name } = event.target;
+    setErrors(prevErrors => ({ ...prevErrors, [name]: '' }));
   };
 
-  handleForgotPasswordLinkClickEvent = () => {
+  const handleForgotPasswordLinkClick = () => {
```
syedsajjadkazmii, line 133: if we are only sending a track event here, can we rename it to `trackForgotPasswordLinkClick`?

src/login/LoginPage.jsx
zainab-amir: Not all handleSubmit requests call `loginRequest`, some fail on the frontend validations and return before that.

src/login/LoginPage.jsx
zainab-amir: Its on my agenda to discuss it with Wajeeha during standup.

src/login/data/reducers.js
zainab-amir: The action `backupLoginFormBegin` has the type BACKUP_LOGIN_DATA.BEGIN. We are initiating a backup request when we call `backupLoginFormBegin`, it hasn't succeeded yet.

src/login/AccountActivationMessage.jsx
zainab-amir: It was inconsistent with the login form design. The gap between username/email field and banner was too low and different from the LoginFailure gap

syedsajjadkazmii (approved):

PR #897, opened 2023-05-10
refactor: use paragon Autosuggest field for country field

syedsajjadkazmii:
VAN-1415

### Description

use paragon Autosuggest field for country field

#### JIRA

[VAN-1415](https://2u-internal.atlassian.net/browse/VAN-1415)

Locally

### DIFF
--- a/src/register/ConfigurableRegistrationForm.jsx
+++ b/src/register/ConfigurableRegistrationForm.jsx
@@ -54,13 +54,13 @@ const ConfigurableRegistrationForm = (props) => {
   });
 
   const handleOnChange = (event, countryValue = null) => {
-    const { name, type } = event.target;
+    const { name } = event.target;
     let value;
     if (countryValue) {
       value = { ...countryValue };
     } else {
       value = event.target.type === 'checkbox' ? event.target.checked : event.target.value;
-      if (type === 'checkbox') {
+      if (event.target.type === 'checkbox') {
         setFieldErrors(prevErrors => ({ ...prevErrors, [name]: '' }));
       }
     }
--- a/src/register/registrationFields/CountryField.jsx
+++ b/src/register/registrationFields/CountryField.jsx
@@ -1,154 +1,62 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React from 'react';
 
 import { useIntl } from '@edx/frontend-platform/i18n';
-import { Icon, IconButton } from '@edx/paragon';
-import { ExpandLess, ExpandMore } from '@edx/paragon/icons';
+import { FormAutosuggest, FormAutosuggestOption, FormControlFeedback } from '@edx/paragon';
 import PropTypes from 'prop-types';
 
-import { FormGroup } from '../../common-components';
-import {
-  COUNTRY_CODE_KEY, COUNTRY_DISPLAY_KEY, EXPAND_LESS_ICON, EXPAND_MORE_ICON,
-} from '../data/constants';
+import { COUNTRY_CODE_KEY, COUNTRY_DISPLAY_KEY } from '../data/constants';
 import messages from '../messages';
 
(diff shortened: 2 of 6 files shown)


src/register/ConfigurableRegistrationForm.jsx
```diff
@@ -54,13 +54,13 @@ const ConfigurableRegistrationForm = (props) => {
...
   const handleOnChange = (event, countryValue = null) => {
-    const { name, type } = event.target;
+    const { name } = event.target;
     let value;
     if (countryValue) {
       value = { ...countryValue };
     } else {
       value = event.target.type === 'checkbox' ? event.target.checked : event.target.value;
-      if (type === 'checkbox') {
+      if (event.target.type === 'checkbox') {
```
ahtesham-quraish, line 63: event.target.type i think we should destructure  it that would make more sense

src/register/registrationFields/CountryField.jsx
```diff
@@ -1,154 +1,62 @@
...
-           as blur event is having higher priority than click event and restricts the click event.
-          */
-          onMouseDown={(event) => event.preventDefault()}
-        >
-          {countryName.length > 30 ? countryName.substring(0, 30).concat('...') : countryName}
-        </button>
-      );
-    });
+  const handleSelected = (value) => {
+    if (props.onBlurHandler) { props.onBlurHandler({ target: { name: 'country', value } }); }
```
ahtesham-quraish, line 15: please destructure the onBlurHandler from props here too

src/register/tests/RegistrationPage.test.jsx
```diff
@@ -1041,7 +1043,7 @@ describe('RegistrationPage', () => {
       getLocale.mockImplementation(() => ('ar-ae'));
 
       const registrationPage = mount(reduxWrapper(<IntlRegistrationPage {...props} />));
-      registrationPage.find('input#country').simulate('focus');
+      registrationPage.find('input[name="country"]').simulate('click');
```
ahtesham-quraish, line 1046: Can you please explain why do we need to change the event type ? is it because of paragon dropdown uses?

src/register/ConfigurableRegistrationForm.jsx
syedsajjadkazmii: Yes, you are right! Destructuring is always best but in this use case. We are using this `handleOnChange` function for two types of fields, `text` (for country) and `check box` (for TOS, honor code, etc). 
By using the `onChange` method of the autosuggest field, we get the value only, not the event. means there is no `target.type` available. 
So we are left with two options here, either don't destructure the property that is not available for all use cases or pass hardcoded `type='text'` while calling this `handleOnChange` function. And from these two options, i thought the later one is better. Please let me know if you have any suggestions regarding this.

src/register/registrationFields/CountryField.jsx
syedsajjadkazmii: The reason I did not destructure the handlers here was the naming of handlers and functions of components. the `onFocusHandler` and `onChangeHandler` are of the same names. So, I just kept all handlers inside the `props`. 
If we are to destructure them from props, we need to change the names of component functions.

But if you still feel like its worth destructing and we should change the component's function names. I will be happy to do that.

src/register/tests/RegistrationPage.test.jsx
syedsajjadkazmii: Sorry for misunderstanding the question before^.

Yes, you are right. It's because of the paragon component. They open the dropdown on `onClick` event.

src/register/registrationFields/CountryField.jsx
ahtesham-quraish: its fine we can keep it

ahtesham-quraish (approved):

PR #905, opened 2023-05-16
fix: tests and lint issues

zainab-amir:
### Description

Include a description of your changes here, along with a link to any relevant Jira tickets and/or Github issues.

#### JIRA

[XXX-XXXX](https://2u-internal.atlassian.net/browse/XXX-XXXX)

Include a link to the sandbox for design changes or screenshot for before and after. **Remove this section if its not applicable.**

|Before|After|
|-------|-----|
|      |      |

#### Merge Checklist

#### Post-merge Checklist


src/login/LoginPage.jsx
```diff
@@ -229,6 +229,7 @@ const LoginPage = (props) => {
             name="password"
             value={formFields.password}
             autoComplete="off"
+            showScreenReaderText={false}
```
syedsajjadkazmii, line 232: For my understanding, why is this prop introduced?

src/login/data/reducers.js
```diff
@@ -47,6 +47,7 @@ const reducer = (state = defaultState, action = {}) => {
         loginResult: action.payload,
       };
     case LOGIN_REQUEST.FAILURE:
+      // eslint-disable-next-line no-case-declarations
```
syedsajjadkazmii, line 50: Having variables in case like this cause the variable being out of scope of this particular case. Therefore it is discouraged.

instead of disabling the eslint here, can we do like this?

```
    case LOGIN_REQUEST.FAILURE: {
      const { email, loginError, redirectUrl } = action.payload;
      return {
        ...state,
        loginErrorCode: loginError.errorCode,
        loginErrorContext: { ...loginError.context, email, redirectUrl },
        submitState: DEFAULT_STATE,
      };
    }

zainab-amir: How is it different from the code I have already written?

src/login/LoginPage.jsx
zainab-amir: The sr-only message ("Password must contain at least 8 characters, at least one letter, and at least one number". ) on the password field was written considering the tooltip in mind that appears on the registration page. The tooltip is not accessible for screen readers and they need a way to see the password requirements. 

This message is however not needed on the login page.

syedsajjadkazmii (approved):

PR #913, opened 2023-05-18
fix: bugs in country field

syedsajjadkazmii:
VAN-1415

### Description

This PR fixed a bug in the country field.
**Bug:** 
Validations were being run when drop-down arrow was clicked. Which should not run.

#### JIRA

[VAN-1415](https://2u-internal.atlassian.net/browse/VAN-1415)

Include a link to the sandbox for design changes or screenshots before and after.

**Before**

https://github.com/openedx/frontend-app-authn/assets/52817156/8ce60555-3007-4408-8eaf-3d4d491cada3

**After**

https://github.com/openedx/frontend-app-authn/assets/52817156/b053725f-39ec-4025-97b0-02872819fd4d

### DIFF
--- a/src/register/registrationFields/CountryField.jsx
+++ b/src/register/registrationFields/CountryField.jsx
@@ -15,6 +15,14 @@ const CountryField = (props) => {
     if (props.onBlurHandler) { props.onBlurHandler({ target: { name: 'country', value } }); }
   };
 
+  const onBlurHandler = (event) => {
+    // Do not run validations when drop-down arrow is clicked
+    if (event.relatedTarget && event.relatedTarget.className.includes('pgn__form-autosuggest__icon-button')) {
+      return;
+    }
+    if (props.onBlurHandler) { props.onBlurHandler(event); }
+  };
+
   const onFocusHandler = (event) => {
     if (props.onFocusHandler) { props.onFocusHandler(event); }
   };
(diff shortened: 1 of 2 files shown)


src/register/tests/RegistrationPage.test.jsx
```diff
@@ -1326,5 +1326,16 @@ describe('RegistrationPage', () => {
       expect(registrationPage.find('div.alert-heading').length).toEqual(1);
       expect(registrationPage.find('div.alert').first().text()).toContain('An error occured');
     });
+
+    it('should not run country field validation when onBlur is fired by drop-down error click', () => {
```
zainab-amir, line 1330:
```suggestion
    it('should not run country field validation when onBlur is fired by drop-down arrow icon click', () => {
```

src/register/registrationFields/CountryField.jsx
```diff
@@ -15,6 +15,14 @@ const CountryField = (props) => {
     if (props.onBlurHandler) { props.onBlurHandler({ target: { name: 'country', value } }); }
   };
 
+  const onBlurHandler = (event) => {
+    // Do not run validations when drop-down arrow is clicked
+    if (event.relatedTarget && event.relatedTarget.className.includes('pgn__form-autosuggest__icon-button')) {
+      return;
+    }
+    if (props.onBlurHandler) { props.onBlurHandler(event); }
```
zainab-amir, line 23: Should we rename the `props.onBlurHandler` to something different as its being used in both on blur and on click events?

syedsajjadkazmii: @zainab-amir Yes, you are right, it's being used for both `onBlur` and `onSelected` and the reason is we want to validate the input on both `onBlur` and `onSelected` events. And in authn, we know that validations are run by the `onBlur` event. Therefore I think it makes sense. 

But if you have suggestions for a name. please let me know.

zainab-amir (approved):

PR #936, opened 2023-06-06
feat: upgrade react router to v6

Syed-Ali-Abbas-Zaidi:
## Ticket
[React Router Upgrade to v6](https://github.com/openedx/platform-roadmap/issues/276).

## Description
This PR upgrades React Router from `v5` to `v6`.


mubbsharanwar (approved):
Looks good.

src/common-components/RedirectLogistration.jsx
```diff
@@ -49,13 +47,13 @@ const RedirectLogistration = (props) => {
       }
       const registrationResult = { redirectUrl: finalRedirectUrl, success };
       return (
-        <Redirect to={{
-          pathname: AUTHN_PROGRESSIVE_PROFILING,
-          state: {
+        <Navigate
+          to={AUTHN_PROGRESSIVE_PROFILING}
+          state={{
```
mubbsharanwar, line 52: The location state is not setting which causes the flow break of the welcome page. This means the user is unable to fill in data on the progressive profiling/welcome page. 
This state is used in [ProgressiveProfilling](https://github.com/openedx/frontend-app-authn/blob/dff3903617fe96fce71197427b75155311af8fe9/src/progressive-profiling/ProgressiveProfiling.jsx#L82) component which we are setting here in RedirectLogistration component.

mubbsharanwar (changes requested):
Need to fix setting the state on the location object of react-dom.

mubbsharanwar (approved):

PR #937, opened 2023-06-06
fix: remove survey cookie

attiyaIshaque:
### Description

Remove survey cookie from authn

#### JIRA

[VAN-1429](https://2u-internal.atlassian.net/browse/VAN-1429)


attiyaIshaque:
> Can you please add comment in our code why we need `REGISTER_CONVERSION_COOKIE_NAME` [cookie](https://github.com/openedx/frontend-app-authn/blob/master/src/register/RegistrationPage.jsx#L227-L230) here in RegistrationPage for feature reference.

@mubbsharanwar Addressed your comment.

mubbsharanwar (approved):

PR #940, opened 2023-06-13
Ahtesham/van van 1474/redirect user to welcome page

ahtesham-quraish:
Description
Redirect the user to welcome page after registration, as Authn is embedded in iframe which is located in Hubspot so we call window.postMessage function from Authn which let the Hubspot knows about redirect URL and then reload the welcome page in parent window.

How Has This Been Tested?
this is tested locally but when live site will be ready then we can test there as well.

Its related to : https://2u-internal.atlassian.net/browse/VAN-1474


ahtesham-quraish:
no need

PR #942, opened 2023-06-13
feat: registration conversion events

mubbsharanwar:
Passing host property to registration and progressive profiling events to calculate on-ramp conversion rate.

VAN-1478

### DIFF
--- a/src/progressive-profiling/ProgressiveProfiling.jsx
+++ b/src/progressive-profiling/ProgressiveProfiling.jsx
@@ -142,6 +142,8 @@ const ProgressiveProfiling = (props) => {
     return null;
   }
 
+  const queryParams = getAllPossibleQueryParams();
+
   const handleSubmit = (e) => {
     e.preventDefault();
     window.history.replaceState(location.state, null, '');
@@ -162,6 +164,7 @@ const ProgressiveProfiling = (props) => {
         isGenderSelected: !!values.gender,
         isYearOfBirthSelected: !!values.year_of_birth,
         isLevelOfEducationSelected: !!values.level_of_education,
+        host: queryParams?.host || '',
       },
     );
   };
@@ -170,7 +173,12 @@ const ProgressiveProfiling = (props) => {
     e.preventDefault();
     window.history.replaceState(props.location.state, null, '');
     setShowModal(true);
-    sendTrackEvent('edx.bi.welcome.page.skip.link.clicked');
+    sendTrackEvent(
+      'edx.bi.welcome.page.skip.link.clicked',
+      {
+        host: queryParams?.host || '',
+      },
+    );
   };
 
   const onChangeHandler = (e) => {
--- a/src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
+++ b/src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
@@ -12,7 +12,11 @@ import { MemoryRouter, Router } from 'react-router-dom';
 import configureStore from 'redux-mock-store';
 
 import {
-  COMPLETE_STATE, DEFAULT_REDIRECT_URL, EMBEDDED, FAILURE_STATE, RECOMMENDATIONS,
+  AUTHN_PROGRESSIVE_PROFILING,
+  COMPLETE_STATE, DEFAULT_REDIRECT_URL,
+  EMBEDDED,
+  FAILURE_STATE,
+  RECOMMENDATIONS,
 } from '../../data/constants';
 import { saveUserProfile } from '../data/actions';
 import ProgressiveProfiling from '../ProgressiveProfiling';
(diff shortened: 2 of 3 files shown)


src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -126,6 +136,7 @@ const ProgressiveProfiling = (props) => {
         isGenderSelected: !!values.gender,
         isYearOfBirthSelected: !!values.year_of_birth,
         isLevelOfEducationSelected: !!values.level_of_education,
+        variant: embeddedVariant,
```
syedsajjadkazmii, line 139: this will add `variant: ''` for the non-embedded experience. I think we don't want to add this property for the non-embedded experience.

syedsajjadkazmii: this concern is valid for other events as well.

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -103,8 +109,12 @@ const ProgressiveProfiling = (props) => {
...
   }
 
-  const optionalFields = location.state.optionalFields.fields;
-  const extendedProfile = location.state.optionalFields.extended_profile;
+  let optionalFields = {};
+  let extendedProfile = [];
+  if (location?.state) {
+    optionalFields = location.state.optionalFields.fields;
+    extendedProfile = location.state.optionalFields.extended_profile;
+  }
```
syedsajjadkazmii, line 117: why are you doing this? that's not the part of this ticket.

src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
```diff
@@ -151,12 +151,51 @@ describe('ProgressiveProfilingTests', () => {
     expect(store.dispatch).toHaveBeenCalledWith(saveUserProfile('abc123', formPayload));
   });
 
+  it('should fire event with on ramp experience on form submit click button', async () => {
```
syedsajjadkazmii, line 154: thoughts? 
``` suggestion
  it('should add variant property in submit event for embedded experience', async () => {
```

src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
```diff
@@ -151,12 +151,51 @@ describe('ProgressiveProfilingTests', () => {
...
+      variant: 'embedded',
+    };
+    delete window.location;
+    window.location = { href: getConfig().BASE_URL.concat(AUTHN_PROGRESSIVE_PROFILING), search: '?variant=embedded' };
+    const progressiveProfilingPage = await getProgressiveProfilingPage();
+    progressiveProfilingPage.find('button.btn-brand').simulate('click');
+    expect(sendTrackEvent).toHaveBeenCalledWith('edx.bi.welcome.page.submit.clicked', eventProperties);
+  });
+
+  it('should fire event with normal register form on form submit click button', async () => {
```
syedsajjadkazmii, line 168: if we are not going to add this property for non-embedded experience, then
``` suggestion
  it('should not add variant property in submit event for embedded experience', async () => {
```

src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
```diff
@@ -151,12 +151,51 @@ describe('ProgressiveProfilingTests', () => {
...
+      variant: '',
+    };
+    delete window.location;
+    window.location = { href: getConfig().BASE_URL.concat(AUTHN_PROGRESSIVE_PROFILING) };
+    const progressiveProfilingPage = await getProgressiveProfilingPage();
+    progressiveProfilingPage.find('button.btn-brand').simulate('click');
+    expect(sendTrackEvent).toHaveBeenCalledWith('edx.bi.welcome.page.submit.clicked', eventProperties);
+  });
+
+  it('should fire event with embedded experience on pressing skip for now button', async () => {
```
syedsajjadkazmii, line 182:
```suggestion
  it('should add variant property in skip event for embedded experience, async () => {
```

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: Does empty value affect our stats? I think it is fine if we are passing the '' value for non or ramp experience. We can easily differentiate the traffic on the welcome page by passing this property whatever the value would be.

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: By registering through the non-embedded experience we set this location state and use this state for progressive profiling, but through iframe embedded experience we are unable to set the location state that's why we need to handle this case when we are not setting the state.

zainab-amir: This will be handled in my PR.

src/progressive-profiling/ProgressiveProfiling.jsx
syedsajjadkazmii: yes, you are right. but if we are never going to use this then why pass it?

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -70,6 +71,11 @@ const ProgressiveProfiling = (props) => {
   }, [DASHBOARD_URL, registrationResponse]);
 
   useEffect(() => {
+    const queryParams = getAllPossibleQueryParams();
+    if (authenticatedUser?.userId && queryParams?.host) {
```
syedsajjadkazmii, line 75: do you think there should be a check for embedded too?

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -70,6 +71,11 @@ const ProgressiveProfiling = (props) => {
...
+      setEmbeddedHost(queryParams?.host);
```
zainab-amir, line 76: Do we need to keep a separate state? We can just simply use the query param constant directly in the events i.e
```
sendTrackEvent(
      'edx.bi.welcome.page.skip.link.clicked',
      {
        host: queryParams?.host,
      },
    );
```

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: no.

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: We can add but there is no need to add this because `host` is also fullfling the same purpose.

src/progressive-profiling/ProgressiveProfiling.jsx
zainab-amir: We already decided that its okay if its empty i.e user is coming from a normal flow.

mubbsharanwar: I was talking about the embedded flow to set event property before firing the event.

syedsajjadkazmii (approved):

PR #944, opened 2023-06-15
Allow welcome page to load for embedded experience

zainab-amir:
### Description

- Updated progressive profiling to call the backend API if the query param `variant=embedded` is available

Cases covered:
- It should not render form for unauthenticated users
- It should render welcome page without query params only if its coming from the normal registration flow
  - It should redirect user to the redirectURL available in the registration response
  - It should not make call to the backend API in this case
 - It should render welcome page with query params by calling the backend API to get the fields
   - It should redirect user to `SEARCH_CATALOG_URL`

#### JIRA

[VAN-1472](https://2u-internal.atlassian.net/browse/VAN-1472)

Added unit tests


syedsajjadkazmii (approved):

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -39,47 +40,78 @@ import {
...
-    formRenderState, submitState, showError, location,
+    getFieldDataFromBackend,
+    location,
+    submitState,
+    showError,
+    welcomePageContext,
+    welcomePageContextApiStatus,
   } = props;
+  const queryParams = useMemo(() => getAllPossibleQueryParams(), []);
+  const isRegistrationEmbedded = queryParams?.variant === EMBEDDED;
```
syedsajjadkazmii, line 53: please use this util created https://github.com/openedx/frontend-app-authn/blob/d41c06b1fd048572342f05e38af84267e7df82b2/src/data/utils/dataUtils.js#L80

syedsajjadkazmii (approved):

PR #946, opened 2023-06-15
fix: avoid validations on blur event

ahtesham-quraish:
Description:

When Authn is embedded in iframe then validations should not runs on blur event it should only run when button is clicked.

Its related to : [VAN-1481](https://2u-internal.atlassian.net/browse/VAN-1481)

Its has been tested locally.

### DIFF
--- a/src/register/ConfigurableRegistrationForm.jsx
+++ b/src/register/ConfigurableRegistrationForm.jsx
@@ -34,6 +34,7 @@ const ConfigurableRegistrationForm = (props) => {
     setFieldErrors,
     setFocusedField,
     setFormFields,
+    registrationEmbedded,
   } = props;
 
   let showTermsOfServiceAndHonorCode = false;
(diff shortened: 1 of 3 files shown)


src/register/ConfigurableRegistrationForm.jsx
```diff
@@ -77,6 +78,9 @@ const ConfigurableRegistrationForm = (props) => {
       const { countryCode, displayValue } = countryValidation;
       error = countryValidation.error;
       setFormFields(prevState => ({ ...prevState, country: { countryCode, displayValue } }));
+      if (isRegistrationEmbedded) {
+        return;
```
syedsajjadkazmii, line 82: 1. why are you adding this at this place of code?
2. why are we restricting it to not run onBlur for country field only? there can be other configurable fields as well and rest of the else blocks are still validation them.

mubbsharanwar (approved):

PR #950, opened 2023-06-20
feat: keep username suggestion for origin

zainab-amir:
### Description

For the embedded registration form experience even though we do not want to show error on blur we still want to keep the username suggestions that are fetched on the "name" field on blur event. 

#### JIRA

[VAN-1498](https://2u-internal.atlassian.net/browse/VAN-1498)

**Don't show error on blur for empty "name" field** 

https://github.com/openedx/frontend-app-authn/assets/40633976/56a30c64-3712-4fa1-af1c-fe4dfa7826c8

**Don't show error on blur for invalid "name" field**

https://github.com/openedx/frontend-app-authn/assets/40633976/88f57239-9aa5-4994-abb5-18ebf69ca7a8

**Show username suggestion**

https://github.com/openedx/frontend-app-authn/assets/40633976/2f86cda3-91e5-4e9c-bb7d-8ee0dfeec61a

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -412,10 +412,19 @@ const RegistrationPage = (props) => {
   };
 
   const handleOnBlur = (event) => {
+    const { name, value } = event.target;
     if (registrationEmbedded) {
+      if (name === 'name') {
+        validateInput(
+          name,
+          value,
+          { name: formFields.name, username: formFields.username, form_field_key: name },
+          !validationApiRateLimited,
+          false,
+        );
+      }
       return;
     }
-    const { name, value } = event.target;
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -412,10 +412,19 @@ const RegistrationPage = (props) => {
   };
 
   const handleOnBlur = (event) => {
+    const { name, value } = event.target;
     if (registrationEmbedded) {
```
syedsajjadkazmii, line 416: can't we do something like
```suggestion
    if (registrationEmbedded && name != 'name') {
```

syedsajjadkazmii (approved):

PR #952, opened 2023-06-21
fix: hide Zendesk widget for embedded experience

syedsajjadkazmii:
### Description

This Pr adds a check to hide Zendesk widget for embedded experience

### DIFF
--- a/src/common-components/Zendesk.jsx
+++ b/src/common-components/Zendesk.jsx
@@ -5,6 +5,7 @@ import { useIntl } from '@edx/frontend-platform/i18n';
 import Zendesk from 'react-zendesk';
 
 import messages from './messages';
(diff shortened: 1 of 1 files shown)


src/common-components/Zendesk.jsx
```diff
@@ -48,6 +50,10 @@ const ZendeskHelp = () => {
     },
   };
 
+  if (window.location.pathname === REGISTER_EMBEDDED_PAGE && isHostAvailableInQueryParams()) {
```
zainab-amir, line 53: I think we can make do without the `isHostAvailableInQueryParams`. The page embedded registration is only ever accessible if host is present.

syedsajjadkazmii: yes you are right

zainab-amir (approved):

PR #958, opened 2023-06-27
refactor: base component name and folder structure

zainab-amir:
### Description

As part of VAN-1506, we want to A/B test a different sidebar. The current folder structure for base component did not support that. In this ticket, I have updated the structure for the base component and renamed it to be more conventional.  

#### JIRA

[VAN-1506](https://2u-internal.atlassian.net/browse/VAN-1506)

By running already existing tests


mubbsharanwar (commented):
Why we are renaming it to the `left-sidebar`? Technically it's not the left sidebar. Left sidebar is inside the `LargeLayout` and `AuthLargeLayout` components.

mubbsharanwar (approved):

PR #993, opened 2023-07-21
feat: param based redirection

mubbsharanwar:
Register success redirection would be based on the next query params.

VAN-1535

### DIFF
--- a/src/common-components/data/reducers.js
+++ b/src/common-components/data/reducers.js
@@ -3,7 +3,10 @@ import { COMPLETE_STATE, FAILURE_STATE, PENDING_STATE } from '../../data/constan
 
 export const defaultState = {
   fieldDescriptions: {},
-  optionalFields: {},
+  optionalFields: {
+    fields: {},
+    extended_profile: [],
+  },
   thirdPartyAuthApiStatus: null,
   thirdPartyAuthContext: {
     currentProvider: null,
@@ -13,6 +16,7 @@ export const defaultState = {
     secondaryProviders: [],
     pipelineUserDetails: null,
     errorMessage: null,
+    welcomePageRedirectUrl: null,
   },
 };
 
--- a/src/progressive-profiling/ProgressiveProfiling.jsx
+++ b/src/progressive-profiling/ProgressiveProfiling.jsx
@@ -23,12 +23,12 @@ import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 
 import { saveUserProfile } from './data/actions';
+import { welcomePageContextSelector } from './data/selectors';
 import messages from './messages';
 import ProgressiveProfilingPageModal from './ProgressiveProfilingPageModal';
 import BaseContainer from '../base-container';
 import { RedirectLogistration } from '../common-components';
 import { getThirdPartyAuthContext } from '../common-components/data/actions';
-import { optionalFieldsSelector } from '../common-components/data/selectors';
 import {
   COMPLETE_STATE,
   DEFAULT_REDIRECT_URL,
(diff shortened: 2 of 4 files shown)


src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -171,6 +172,9 @@ const ProgressiveProfiling = (props) => {
 
   const handleSkip = (e) => {
     e.preventDefault();
+    if (nextUrl) {
+      window.location = nextUrl;
```
zainab-amir, line 176: why do we need to add it here?

zainab-amir (commented):
Why are we not using existing logic in use effect where we set the optional fields?

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: We also need to redirect to the next URL for the skip button click.

src/common-components/data/reducers.js
```diff
@@ -3,7 +3,10 @@ import { COMPLETE_STATE, FAILURE_STATE, PENDING_STATE } from '../../data/constan
 
 export const defaultState = {
   fieldDescriptions: {},
-  optionalFields: {},
+  optionalFields: {
+    fields: [],
+    extended_profile: [],
+  },
```
zainab-amir, line 9: Why have we changed the default state?

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -90,17 +91,18 @@ const ProgressiveProfiling = (props) => {
 
   useEffect(() => {
     if (registrationEmbedded) {
-      getFieldDataFromBackend({ is_welcome_page: true });
+      getFieldDataFromBackend({ is_welcome_page: true, next: queryParams?.next });
     }
-  }, [registrationEmbedded, getFieldDataFromBackend]);
+  }, [registrationEmbedded, getFieldDataFromBackend, queryParams?.next]);
```
zainab-amir, line 96: We can use queryParams directly
```suggestion
  }, [registrationEmbedded, getFieldDataFromBackend, queryParams]);
```

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -172,7 +172,9 @@ const ProgressiveProfiling = (props) => {
   const handleSkip = (e) => {
     e.preventDefault();
     window.history.replaceState(props.location.state, null, '');
-    setShowModal(true);
+    if (welcomePageContext.nextUrl !== '') {
+      setShowModal(true);
+    }
```
zainab-amir, line 177: Why are we changing the functionality?

mubbsharanwar: Modal shows the button, click to redirect to the dashboard but in the embedded experience we want to redirect to the next URL.

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: We can't use objects directly in dependency. It causes the infinite useEffect loop due to memory address change.

syedsajjadkazmii: @mubbsharanwar the useEffect should not run infinite when having object as dependency, it must be some other reason that useEffect is running infinitely for queryParams.

zainab-amir: Yes. The queryParams are constant and will only change once

src/progressive-profiling/ProgressiveProfiling.jsx
zainab-amir: Then lets change the text of the button. We don't want to change functionality without asking product and UX

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: Why should not we use objects as a dependency? https://dmitripavlutin.com/react-useeffect-infinite-loop/#21-avoid-objects-as-dependencies

src/progressive-profiling/ProgressiveProfiling.jsx
mubbsharanwar: Actually, I didn't set the env variables locally that's why I was getting the label button on the modal button "Go to dashboard". On stage and prod the modal button label is "Go to edX" which is fine. We don't need to add this check.

zainab-amir (approved):

PR #1001, opened 2023-07-26
feat: Add segments events

mubbsharanwar:
Add segments events to track popular and trending recommendations.

[VAN-1569](https://2u-internal.atlassian.net/browse/VAN-1569)

### DIFF
--- a/src/recommendations/ProductCard/index.jsx
+++ b/src/recommendations/ProductCard/index.jsx
@@ -9,8 +9,7 @@ import { createCodeFriendlyProduct, getVariant, useProductType } from '../data/u
 import {
   cardBadgesMessages,
 } from '../messages';
-import { trackRecommendationCardClickOptimizely } from '../optimizelyExperiment';
-import { trackRecommendationsClicked } from '../track';
+import { trackRecommendationClick } from '../track';
 
 const ProductCard = ({
   product,
@@ -64,20 +63,16 @@ const ProductCard = ({
     ],
   );
   const handleCardClick = () => {
-    trackRecommendationCardClickOptimizely(userId?.toString());
-    trackRecommendationsClicked(
-      product.courseKey,
-      false,
+    trackRecommendationClick(
+      product,
       position + 1,
+      false,
       userId,
-      product.marketingUrl,
-      product.recommendationType || 'algolia',
     );
   };
 
   return (
     <BaseCard
-      url={product.url}
       customHeaderImage={headerImage}
       schoolLogo={isMultipleOwner ? '' : schoolLogo}
       title={product.title}
--- a/src/recommendations/track.js
+++ b/src/recommendations/track.js
@@ -3,7 +3,7 @@ import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 export const LINK_TIMEOUT = 300;
 
 export const eventNames = {
-  recommendedCourseClicked: 'edx.bi.user.recommended.course.click',
+  recommendedProductClicked: 'edx.bi.user.recommended.product.clicked',
   recommendationsGroup: 'edx.bi.user.recommendations.group',
   recommendationsViewed: 'edx.bi.user.recommendations.viewed',
 };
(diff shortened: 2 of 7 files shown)


src/recommendations/track.js
```diff
@@ -17,13 +17,17 @@ export const createLinkTracker = (tracker, href, openInNewTab = false) => (e) =>
   return setTimeout(() => { global.location.href = href; }, LINK_TIMEOUT);
 };
 
-export const trackRecommendationsClicked = (courseKey, isControl, position, userId, href, recommendationType) => {
+export const trackRecommendationsClicked = (
```
syedsajjadkazmii, line 20: can we please rename it?
```suggestion
export const trackRecommendationClick = (
```

src/recommendations/track.js
```diff
@@ -32,11 +36,12 @@ export const trackRecommendationsClicked = (courseKey, isControl, position, user
...
 
-export const trackRecommendationsViewed = (recommendedCourseKeys, isControl, userId) => {
+export const trackRecommendationsViewed = (recommendedProducts, isControl, userId, type) => {
   sendTrackEvent(
     eventNames.recommendationsViewed, {
       page: 'authn_recommendations',
-      course_key_array: recommendedCourseKeys,
+      recommendation_type: type,
+      products: recommendedProducts,
       amplitude_recommendations: false,
```
syedsajjadkazmii, line 45: i think is extra property

src/recommendations/RecommendationsList.jsx
```diff
@@ -15,6 +16,17 @@ const RecommendationsList = (props) => {
     title, recommendations, userId, setSelectedRecommendationsType, selectedRecommendationsType,
   } = props;
 
+  useEffect(() => {
+    if (recommendations.length) {
+      const viewedProductsList = recommendations.map((product) => ({
```
syedsajjadkazmii, line 21: this code should be part of the `trackRecommendationsViewed` function. We should simply pass products here.

src/recommendations/RecommendationsList.jsx
```diff
@@ -15,6 +16,17 @@ const RecommendationsList = (props) => {
...
+        product_key: product.uuid,
+        product_line: product.cardType,
+        product_source: product.productSource.name,
+      }));
+      trackRecommendationsViewed(viewedProductsList, false, userId);
+    }
+  }, [recommendations, userId]);
```
syedsajjadkazmii, line 28: `selectedRecommendationsType` should also be part of this useEffect. otherwise on tab change, this event will not be fired.

src/recommendations/RecommendationsList.jsx
```diff
@@ -15,6 +16,17 @@ const RecommendationsList = (props) => {
     title, recommendations, userId, setSelectedRecommendationsType, selectedRecommendationsType,
   } = props;
 
+  useEffect(() => {
```
syedsajjadkazmii, line 19: Why is this hook not in the `recommendationsPage` component? According to our approach, all setup-related things are in the `recommendationsPage` component, and the `recommendationsList` component's responsibility is to simply render and display recommendations.

src/recommendations/ProductCard/index.jsx
```diff
@@ -75,11 +74,12 @@ const ProductCard = ({
     ],
   );
   const handleCardClick = () => {
-    trackRecommendationCardClickOptimizely(userId?.toString());
     trackRecommendationsClicked(
-      product.courseKey,
-      false,
+      product.uuid,
```
syedsajjadkazmii, line 78: does `uuid` contain `productKey` for all products? I think `uuid` is just for programs, what about courses, and other types?

src/recommendations/ProductCard/index.jsx
```diff
@@ -75,11 +74,12 @@ const ProductCard = ({
...
+      product.cardType,
```
syedsajjadkazmii, line 79: does cardType always contain the correct type? is it confirmed?

src/recommendations/ProductCard/index.jsx
```diff
@@ -75,11 +74,12 @@ const ProductCard = ({
...
+      product.productSource.name,
       position + 1,
+      false,
```
syedsajjadkazmii, line 82: [Confirming] isControl=false means user is seeing the recs?

syedsajjadkazmii (commented):
please write tests to complete the coverage.

src/recommendations/ProductCard/index.jsx
mubbsharanwar: Yes.

src/recommendations/ProductCard/index.jsx
mubbsharanwar: Yes, It will be always.

src/recommendations/ProductCard/index.jsx
mubbsharanwar: Currently Its is in all products but I write a function `generateProductKey` to make product key according to product.

src/recommendations/RecommendationsList.jsx
mubbsharanwar: Addressed.

src/recommendations/RecommendationsList.jsx
mubbsharanwar: Addressed.

src/recommendations/track.js
mubbsharanwar: Removed

syedsajjadkazmii (approved):

PR #1024, opened 2023-08-08
feat: fetch algolia recommendations

syedsajjadkazmii:
### Description

- Adds logic to fetch Algolia recommendations for post registration recommendations

#### JIRA

[VAN-1599](https://2u-internal.atlassian.net/browse/VAN-1599)

Locally


src/recommendations/data/hooks/useAlgoliaRecommendations.jsx
```diff
@@ -0,0 +1,85 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    if (!functionalCookiesEnabled) {
+      setIsLoading(false);
+      return;
+    }
+
+    const getSearchFiltersQueryString = () => {
+      if (userCountry) {
```
mubbsharanwar, line 36: Please move this condition inside `getLocationRestrictionFilter` function.

src/data/algolia.js
```diff
@@ -0,0 +1,23 @@
...
+const initializeSearchClient = () => algoliasearch(
+  getConfig().ALGOLIA_APP_ID,
+  getConfig().ALGOLIA_SEARCH_API_KEY,
+);
+
+const getLocationRestrictionFilter = (userCountry) => (
+  `NOT blocked_in:"${userCountry}" AND (allowed_in:"null" OR allowed_in:"${userCountry}")`
+);
+
+const getPrecisionRadius = () => METERS_IN_ONE_MILE * PRECISION_RADIUS_MILES;
```
mubbsharanwar, line 17: How are we setting precision criteria? Have we any certain requirement or ask?

src/recommendations/data/hooks/useStaticRecommendations.jsx
```diff
@@ -4,7 +4,12 @@ import { getConfig } from '@edx/frontend-platform';
...
 import { filterLocationRestriction } from '../utils';
 
-export default function useProducts(countryCode) {
+/**
+ * This hooks returns Static recommendations (Trending and Popular products) after applying location restrictions.
+ * @param countryCode
+ * @returns {{isLoading: boolean, trendingProducts: *[], popularProducts: *[]}}
+ * */
+const useStaticRecommendations = (countryCode) => {
   const [isLoading, setLoading] = useState(true);
```
mubbsharanwar, line 13: Do we need `isLoading` check in this hook?

src/recommendations/data/tests/hooks.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+      );
+
+      expect(result.current.recommendations).toEqual([]);
+      expect(result.current.isLoading).toBe(false);
+    });
+  });
+
+  describe('useStaticRecommendations Tests', () => {
+    it('should return trending and popular products list', async () => {
+      jest.spyOn(JSON, 'parse').mockReturnValue(mockedAlgoliaCourses);
```
mubbsharanwar, line 63: mocked data object `mockedAlgoliaCourses`  name should be generic or should have a separate object for static recommendations.

src/recommendations/data/algoliaResultsParser.js
```diff
@@ -0,0 +1,33 @@
...
+  const camelCasedResult = camelCaseObject(searchResultCourse);
+
+  return {
+    activeCourseRun: {
+      key: camelCasedResult.activeRunKey,
+      type: camelCasedResult.activeRunType,
+      marketingUrl: camelCasedResult.marketingUrl,
+      minEffort: camelCasedResult.minEffort,
+      maxEffort: camelCasedResult.maxEffort,
+      weeksToComplete: camelCasedResult.weeksToComplete,
```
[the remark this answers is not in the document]
syedsajjadkazmii: @mubbsharanwar No, we don't need these keys, the reason is I am making the code consistent with what the prospectus has. They also don't use these in product card display but the process these keys. 
What do you suggest?

src/recommendations/data/hooks/useAlgoliaRecommendations.jsx
syedsajjadkazmii: IMO opinion it doesn't make sense to move conditions inside the function. The purpose of the function is to get the country code and prepare a restriction filter. It should be our responsibility to make sure to only call this function if we have a country code. 
Second, if we move the condition inside the function, there will be no visibility in the hook that this filter can either exist or not. Hook's implementation will give an impression that country restriction is always getting appended.

What do you suggest?

src/data/algolia.js
syedsajjadkazmii: Nope, i am just replicating what prospectus team use: https://github.com/edx/prospectus/blob/master/src/packages/home-page/components/Recommendations/components/GeoRecommendations.jsx#L45

But your concern is valid. 
@zainab-amir do we need any product decision on this?

src/recommendations/data/hooks/useStaticRecommendations.jsx
syedsajjadkazmii: yes we need it, because the in first render the products list is empty and our recommendations page get redirected to next url.

src/recommendations/data/tests/hooks.test.jsx
syedsajjadkazmii: yes you are right. i will update the name.

src/recommendations/data/algoliaResultsParser.js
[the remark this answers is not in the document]
mubbsharanwar: I think currently we don't have any use case where we use these keys, If needed then we can add these.

src/recommendations/data/hooks/useAlgoliaRecommendations.jsx
mubbsharanwar: I think it should be the function's responsibility to handle exceptions too.

src/data/algolia.js
zainab-amir: We don't need to set precision criteria. We want to keep the same functionality we had before with just a check for functional cookies consent.

mubbsharanwar (approved):

PR #1042, opened 2023-08-22
fix: Post registration feedback changes

mubbsharanwar:
Fix post registration recommendations background color and label

[Figma design](https://www.figma.com/file/nDYPI1KPAtzEOXAwESPlgW/Post-completion-recommendations?type=design&node-id=406-5868&mode=design&t=XEG44uIWVxRmAEn5-0)
<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>
Screenshot 2023-08-22 at 3 42 21 PM
</td>
<td>
Screenshot 2023-08-23 at 11 44 55 AM
</td>
</tr>
</table>

[VAN-1616](https://2u-internal.atlassian.net/browse/VAN-1616)

### DIFF
--- a/src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
+++ b/src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
@@ -225,7 +225,11 @@ describe('ProgressiveProfilingTests', () => {
       const progressiveProfilingPage = await getProgressiveProfilingPage();
 
       expect(progressiveProfilingPage.find('button.btn-brand').text()).toEqual('Next');
+<<<<<<< HEAD
       expect(mockNavigate).toHaveBeenCalledWith(RECOMMENDATIONS);
+=======
+      expect(history.location.pathname).toEqual(RECOMMENDATIONS);
+>>>>>>> b2b0d8c (feat: add events for recommendations (#1039))
     });
 
     it('should not redirect to recommendations page if user is on its way to enroll in a course', async () => {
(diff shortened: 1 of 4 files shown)


syedsajjadkazmii:
According to [figma](https://www.figma.com/file/nDYPI1KPAtzEOXAwESPlgW/Post-completion-recommendations?node-id=341%3A5169&mode=dev), the background color is `Light/700`

src/sass/_recommendations_page.scss
```diff
@@ -1,5 +1,11 @@
 $card-gap: 24px;
 
+.recommendations-container {
```
syedsajjadkazmii, line 3: is this working for all screen types?

mubbsharanwar: The large and medium screens were already working fine and still working fine. The issue was with only a small screen and now it's working fine for all screen sizes.

syedsajjadkazmii: okay, so I will explain my concern in detail. The 100vh means 100% of the viewport height. right? So that means if we apply 100% of the viewport height to a sub-div of the page (which is in our case just the recommendations container) will increase the height of that div to 100vh thus creating useless scroll behavior. 

The 100vh class should be applied to the main div of the page, which contains all the content of the page. which is the `<div className="d-flex flex-column bg-light-200">` div in recommendations page. This will give 100% viewport height to the complete page instead of just a sub-div of the main page. 

But then I anticipate one more problem, our page height is not always 100vh, We have a 3 recommendations list that increases the height of the page from 100vh, so if we apply 100vh, the background color will be applied to only 100vh height part of page. So in that case we should apply a minimum height of 100vh. That will solve the issue everywhere.

we just need to put the `min-vh-100` class in the main div of the page (mentioned above).

IMO this will solve the issue. 

The reason why it is not an issue on larger screens is that the page height on larger screens is greater than 100vh so don't notice the issue there but if we have only one recommendations list, we will start noticing the issue there as well. So applying `min-vh-100` will resolve all these issues.

Please let me know if this creates any further issue.

syedsajjadkazmii (approved):

PR #1043, opened 2023-08-22
feat: remove hydrate call

zainab-amir:
### Description

- Removed hydrate authenticated user call from the progressive profiling page and replaced the authenticated user with the details returned by the registration response
- Updated base container to use the username provided by the registration response. Removed extra call for it.

#### JIRA

[VAN-1580](https://2u-internal.atlassian.net/browse/VAN-1580)

Updated tests

No design changes in this PR.

### DIFF
--- a/src/progressive-profiling/ProgressiveProfiling.jsx
+++ b/src/progressive-profiling/ProgressiveProfiling.jsx
@@ -6,9 +6,7 @@ import { identifyAuthenticatedUser, sendPageEvent, sendTrackEvent } from '@edx/f
 import {
   AxiosJwtAuthService,
   configure as configureAuth,
-  ensureAuthenticatedUser,
   getAuthenticatedUser,
-  hydrateAuthenticatedUser,
 } from '@edx/frontend-platform/auth';
 import { useIntl } from '@edx/frontend-platform/i18n';
 import { getLoggingService } from '@edx/frontend-platform/logging';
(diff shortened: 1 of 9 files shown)


src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -57,27 +55,22 @@ const ProgressiveProfiling = (props) => {
...
   const [registrationResult, setRegistrationResult] = useState({ redirectUrl: '' });
   const [formFieldData, setFormFieldData] = useState({ fields: {}, extendedProfile: [] });
-  const [canViewWelcomePage, setCanViewWelcomePage] = useState(false);
   const [values, setValues] = useState({});
   const [showModal, setShowModal] = useState(false);
   const [showRecommendationsPage, setShowRecommendationsPage] = useState(false);
 
   useEffect(() => {
-    configureAuth(AxiosJwtAuthService, { loggingService: getLoggingService(), config: getConfig() });
-    ensureAuthenticatedUser(DASHBOARD_URL)
```
syedsajjadkazmii, line 73: What of the purpose of ensureAuthenticatedUser was to redirect to provided URL incase the user is not authenticated?
https://github.com/openedx/frontend-platform/blob/master/src/auth/AxiosJwtAuthService.js#L250

How are we handling it now? Is it being handled by the condition below?
```
if (
    !(location.state?.registrationResult || registrationEmbedded)
    || welcomePageContextApiStatus === FAILURE_STATE
    || (welcomePageContextApiStatus === COMPLETE_STATE && !Object.keys(welcomePageContext).includes('fields'))
  ) {
    const DASHBOARD_URL = getConfig().LMS_BASE_URL.concat(DEFAULT_REDIRECT_URL);
    global.location.assign(DASHBOARD_URL);
    return null;
  }
```

If it is getting handled, i would still have a feeling that we should have a check if user is not authenticated, then return to Dashboard.

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -298,6 +286,7 @@ ProgressiveProfiling.propTypes = {
 };
 
 ProgressiveProfiling.defaultProps = {
+  authenticatedUser: {},
```
syedsajjadkazmii, line 289: is should be required, no?

zainab-amir: No, for embedded experience we use the getAuthenticatedUser() to get the details.

src/progressive-profiling/ProgressiveProfiling.jsx
zainab-amir: I will confirm this.

zainab-amir: I checked and the user was not being redirected to login page when they were unauthenticated instead they were able to see an empty welcome page.
This has now been fixed in the latest commit.

syedsajjadkazmii (approved):

syedsajjadkazmii (commented):
Can we please add a test that unauthenticated users should redirect to the dashboard? if it is not already added.

syedsajjadkazmii (approved):

PR #1049, opened 2023-08-29
feat: add registration conversion event

zainab-amir:
### Description

Added a new event on frontend for registration conversion. It is fired when user has successfully registered.

#### JIRA

[VAN-1624](https://2u-internal.atlassian.net/browse/VAN-1624)

Added a new test

NA

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -4,7 +4,7 @@ import React, {
 import { connect } from 'react-redux';
 
 import { getConfig, snakeCaseObject } from '@edx/frontend-platform';
-import { sendPageEvent } from '@edx/frontend-platform/analytics';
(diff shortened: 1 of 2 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -228,6 +228,8 @@ const RegistrationPage = (props) => {
 
   useEffect(() => {
     if (registrationResult.success) {
+      sendTrackEvent('edx.bi.user.account.registered.client', {});
+
       // Optimizely registration conversion event
```
mubbsharanwar, line 233: Should we not remove the optimizely event and mape the newly added event into the optimizely?

zainab-amir: If we do it in the same PR, we will lose data. I will do it as a followup.

mubbsharanwar (approved):

PR #1054, opened 2023-09-05
fix: password validation with correct password value on icon blur

syedsajjadkazmii:
### Description

Reflected password field changes everywhere


syedsajjadkazmii:
the patch is 100% covered. I think we can't do anything about the -0.02%.

ahtesham-quraish (approved):

PR #1058, opened 2023-09-08
fix: recs skip btn event gets cancelled on redirection

syedsajjadkazmii:
### Description

The skip button event gets cancelled due to redirection

#### JIRA

BugBash bug

### DIFF
--- a/src/recommendations/RecommendationsPage.jsx
+++ b/src/recommendations/RecommendationsPage.jsx
@@ -19,7 +19,7 @@ import useAlgoliaRecommendations from './data/hooks/useAlgoliaRecommendations';
 import messages from './messages';
 import RecommendationsLargeLayout from './RecommendationsPageLayouts/LargeLayout';
 import RecommendationsSmallLayout from './RecommendationsPageLayouts/SmallLayout';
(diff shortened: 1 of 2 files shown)


src/recommendations/RecommendationsPage.jsx
```diff
@@ -56,7 +56,7 @@ const RecommendationsPage = () => {
   const handleSkip = (e) => {
     e.preventDefault();
     trackSkipButtonClicked(userId);
-    handleSkipRecommendationPage();
+    setTimeout(() => { handleSkipRecommendationPage(); }, LINK_TIMEOUT);
```
ahtesham-quraish, line 59: why we are using settimeout here? Can we not avoid it?

ahtesham-quraish (changes requested):

src/recommendations/RecommendationsPage.jsx
syedsajjadkazmii: we use setTimeout to give a delay before redirecting if we are firing event also

ahtesham-quraish (approved):

PR #1104, opened 2023-11-15
fix: moved registration tests to specific files.

attiyaIshaque:
### Description

All tests that are not directly related to the registration form functionality have been moved to their respective component-specific test files. 

#### JIRA

[VAN-1628](https://2u-internal.atlassian.net/browse/VAN-1628)

### DIFF
--- a/src/register/RegistrationPage.test.jsx
+++ b/src/register/RegistrationPage.test.jsx
@@ -11,19 +11,16 @@ import { mockNavigate, BrowserRouter as Router } from 'react-router-dom';
 import renderer from 'react-test-renderer';
 import configureStore from 'redux-mock-store';
 
-import RegistrationFailureMessage from './components/RegistrationFailure';
 import {
   backupRegistrationFormBegin,
   clearRegistrationBackendError,
   registerNewUser,
   setUserPipelineDataLoaded,
 } from './data/actions';
-import {
-  FIELDS, FORBIDDEN_REQUEST, INTERNAL_SERVER_ERROR, TPA_AUTHENTICATION_FAILURE, TPA_SESSION_EXPIRED,
-} from './data/constants';
+import { INTERNAL_SERVER_ERROR } from './data/constants';
 import RegistrationPage from './RegistrationPage';
 import {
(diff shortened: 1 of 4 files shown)


src/register/RegistrationPage.test.jsx
```diff
@@ -153,11 +148,7 @@ describe('RegistrationPage', () => {
   };
 
   const ssoProvider = {
```
mubbsharanwar, line 150: Do we still need to pass ssoProvider in `formPayload`? Reference line 216.

src/register/components/RegistrationFailure.test.jsx
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  beforeEach(() => {
+    store = mockStore(initialState);
+    configure({
+      loggingService: { logError: jest.fn() },
+      config: {
+        ENVIRONMENT: 'production',
+        LANGUAGE_PREFERENCE_COOKIE_NAME: 'yum',
+      },
```
mubbsharanwar, line 116: Do we need these configs?

src/register/components/RegistrationFailure.test.jsx
```diff
@@ -0,0 +1,207 @@
...
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('Test Registration Failure', () => {
+    mergeConfig({
+      SHOW_CONFIGURABLE_EDX_FIELDS: true,
+    });
```
mubbsharanwar, line 133: Also what about these?

attiyaIshaque: I have removed these.

src/register/components/RegistrationFailure.test.jsx
attiyaIshaque: Yes. we can Pass ENVIRONMENT undefined.

src/register/RegistrationPage.test.jsx
attiyaIshaque: Nope, we don't need this. Thanks for pointing this out.

PR #1127, opened 2024-01-15
feat: fire event when user change the country field

ahtesham-quraish:
Description: Check how many users update country on registration form after we auto populate it with IP. 

[VAN-1793](https://2u-internal.atlassian.net/browse/VAN-1793)

How Has This Been Tested?
It has been tested locally.

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -70,6 +70,7 @@ const RegistrationPage = (props) => {
     userPipelineDataLoaded,
     submitState,
     validations,
+    IPBasedCountryCode,
   } = useSelector(state => state.register);
 
   const {
@@ -177,8 +178,18 @@ const RegistrationPage = (props) => {
 
       // This is used by the "User Retention Rate Event" on GTM
       setCookie(getConfig().USER_RETENTION_COOKIE_NAME, true);
+
+      // This event will be used to track the number of users who have
+      // changed country field and selected the country other than
+      // IP based country.
+      if (configurableFormFields.country?.countryCode !== IPBasedCountryCode) {
+        sendTrackEvent('edx.bi.user.auto-populatedcountry.changed', {
+          originalCountry: IPBasedCountryCode,
+          selectedCountry: configurableFormFields.country.countryCode,
+        });
+      }
     }
-  }, [registrationResult]);
+  }, [registrationResult, IPBasedCountryCode, configurableFormFields.country]);
 
   const handleOnChange = (event) => {
     const { name } = event.target;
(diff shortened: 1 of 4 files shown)


src/register/data/reducers.js
```diff
@@ -108,6 +110,7 @@ const reducer = (state = defaultState, action = {}) => {
         return {
           ...state,
           backendCountryCode: countryCode,
+          IPBasedCountryCode: countryCode,
```
zainab-amir, line 113: why can't we use `backendCountryCode`? Why do we need a new state for it?

src/register/RegistrationPage.jsx
```diff
@@ -177,8 +178,16 @@ const RegistrationPage = (props) => {
...
+      const { country } = configurableFormFields;
+      if (country?.countryCode !== IPBasedCountryCode) {
+        sendTrackEvent('edx.bi.user.auto-populatedcountry.changed', {
+          originalCountry: IPBasedCountryCode,
+          selectedCountry: country.countryCode,
+        });
+      }
     }
-  }, [registrationResult]);
+  }, [registrationResult, IPBasedCountryCode, configurableFormFields]);
```
zainab-amir, line 190: Just pick country in the dependency array. 
```suggestion
  }, [registrationResult, IPBasedCountryCode, configurableFormFields.country]);
```

src/register/RegistrationPage.jsx
```diff
@@ -177,8 +178,16 @@ const RegistrationPage = (props) => {
 
       // This is used by the "User Retention Rate Event" on GTM
       setCookie(getConfig().USER_RETENTION_COOKIE_NAME, true);
+
```
zainab-amir, line 181: Add a note stating the temporary nature of this code

src/register/data/reducers.js
ahtesham-quraish: this ```backendCountryCode``` is reset when ```BACKUP_REGISTRATION_DATA.BEGIN``` is fired so i did not touch that state I created new one which does not get reset when above action is fired.

src/register/RegistrationPage.jsx
ahtesham-quraish: addressed

src/register/RegistrationPage.jsx
ahtesham-quraish: I have added the note please see if it makes sense

src/register/RegistrationPage.jsx
ahtesham-quraish: I have tried to use it but there is lint error which says ```configurableFormFields``` is not in a dependency array

zainab-amir: Remove line number 185 and directly us the const `configurableFormFields.country` value

ahtesham-quraish: yeah that makes sense

zainab-amir (approved):

PR #1131, opened 2024-01-23
feat: split full name into first name and last name

attiyaIshaque:
### Description

Split full name to first name and last name and add to the registration form, both fields(first name and last name) are required.

#### JIRA

[VAN-1802](https://2u-internal.atlassian.net/browse/VAN-1802)

Tested Locally


zainab-amir (changes requested):

zainab-amir (approved):
Please address the comments before merging. For the changes related to description in messages file, please ensure its checked and updated in all places.

PR #1151, opened 2024-01-29
fix: country field is not populated on page load

ahtesham-quraish:
Description: Country field is not populated when page load VAN-1795

[VAN-1795](https://2u-internal.atlassian.net/browse/VAN-1795)

How Has This Been Tested?
it has been tested locally

After Fix:

https://github.com/openedx/frontend-app-authn/assets/12580995/6f4cf965-65b3-497e-87e6-120e8081b75b

### DIFF
--- a/src/register/RegistrationFields/CountryField/CountryField.test.jsx
+++ b/src/register/RegistrationFields/CountryField/CountryField.test.jsx
@@ -172,7 +172,7 @@ describe('CountryField', () => {
       const dropdownItem = container.querySelector('.dropdown-item');
       fireEvent.click(dropdownItem);
 
-      expect(props.onChangeHandler).toHaveBeenCalledTimes(1);
+      expect(props.onChangeHandler).toHaveBeenCalledTimes(2);
       expect(props.onChangeHandler).toHaveBeenCalledWith(
         { target: { name: 'country' } },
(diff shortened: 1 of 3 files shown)


src/register/RegistrationFields/CountryField/CountryField.test.jsx
```diff
@@ -172,7 +172,7 @@ describe('CountryField', () => {
       const dropdownItem = container.querySelector('.dropdown-item');
       fireEvent.click(dropdownItem);
 
-      expect(props.onChangeHandler).toHaveBeenCalledTimes(1);
+      expect(props.onChangeHandler).toHaveBeenCalledTimes(2);
```
zainab-amir, line 175: Why is it called two times?

ahtesham-quraish: I have changed the logic now when country field component is mounted then this check```!selectedCountry.countryCode``` gets true and call the ```onChangeHandler``` with blank country code and then we trigger click event which set the country field value. so In short ```onChangeHandler``` is called twice one for default setup value and second for when we set the value. I have removed the logic from ```ConfigurableRegistrationForm ``` and put it in ```CountrField``` component

mubbsharanwar (approved):

mubbsharanwar (approved):

PR #1157, opened 2024-02-07
fix: update paragon and country field

ahtesham-quraish:
Description: Update Paragon version and changed the country field implementation as per documentation VAN-1814

[JIRA](https://2u-internal.atlassian.net/browse/VAN-1814)

How Has This Been Tested?

It has been tested locally

|After|
|-----|
  
image


src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -31,6 +31,12 @@ const CountryField = (props) => {
   } = props;
   const { formatMessage } = useIntl();
   const dispatch = useDispatch();
+
+  const countryFieldValue = {
+    userProvidedText: selectedCountry.displayValue,
+    selectionValue: selectedCountry.countryCode,
+    selectionId: selectedCountry.countryCode,
+  };
```
syedsajjadkazmii, line 39: [NIT] Can we add extra/empty line after this?

src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -49,7 +55,7 @@ const CountryField = (props) => {
         { target: { name: 'country' } },
         { countryCode, displayValue: countryDisplayValue },
       );
-    } else if (!selectedCountry.countryCode) {
+    } else if (!selectedCountry.displayValue) {
```
syedsajjadkazmii, line 58: why did we change this?

src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -65,30 +71,27 @@ const CountryField = (props) => {
...
-
   const handleOnFocus = (event) => {
     handleErrorChange('country', '');
     dispatch(clearRegistrationBackendError('country'));
     onFocusHandler(event);
   };
 
   const handleOnChange = (value) => {
-    onChangeHandler({ target: { name: 'country' } }, { countryCode: '', displayValue: value });
+    onChangeHandler({ target: { name: 'country' } }, { countryCode: value.selectionId, displayValue: value.userProvidedText });
```
syedsajjadkazmii, line 87:
```suggestion
    onChangeHandler({ target: { name: 'country' } }, { countryCode: value.selectionValue, displayValue: value.userProvidedText });
```

src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -65,30 +71,27 @@ const CountryField = (props) => {
...
+    if (value.selectionId !== '') {
+      handleOnBlur({ target: { name: 'country', value: value.userProvidedText } });
+    }
```
syedsajjadkazmii, line 90: What scenario is being handled here?

src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -65,30 +71,27 @@ const CountryField = (props) => {
 
     const { value } = event.target;
 
-    const { countryCode, displayValue, error } = validateCountryField(
+    const { error } = validateCountryField(
       value.trim(), countryList, formatMessage(messages['empty.country.field.error']), formatMessage(messages['invalid.country.field.error']),
     );
-
-    onChangeHandler({ target: { name: 'country' } }, { countryCode, displayValue });
```
syedsajjadkazmii, line 72: why is this removed?

syedsajjadkazmii (commented):
(After approval) Can we coordinate with Frontend-BOM regarding this paragon update for additional safeguard?

src/register/RegistrationFields/CountryField/CountryField.jsx
ahtesham-quraish: If we click the dropdown icon and select the item from list that mean we are not focusing the input box we just use the dropdown icon and select the element

src/register/RegistrationFields/CountryField/CountryField.jsx
ahtesham-quraish: Now we are calling this onChangeHandler inside the onChange function and it does not make sense to keep it here inside the blur event

src/register/RegistrationFields/CountryField/CountryField.jsx
ahtesham-quraish: This selectionValue does not have country code it keep display value which is normal country name not its code that is why I have to use sectionId instead of selectionValue

src/register/RegistrationFields/CountryField/CountryField.jsx
ahtesham-quraish: if we put some wrong value inside the country field and we move around the login and signup tabs then this !selectedCountry.countryCode becomes always true and onChangeHandler set the country field with empty data as we are calling this onChangeHandler with blank data so this reset the country field, actual behaviour is to keep whatever user have input in country field should be there in tabs switch

src/register/RegistrationFields/CountryField/CountryField.jsx
syedsajjadkazmii: then for consistency we should also change this (set displayValue in selectionValue)

```

  const countryFieldValue = {
    userProvidedText: selectedCountry.displayValue,
    selectionValue: selectedCountry.displayValue,
    selectionId: selectedCountry.countryCode,
  };
```

src/register/RegistrationFields/CountryField/CountryField.jsx
syedsajjadkazmii: isn't the `selectionId` provided on item click? i wanted to understand this check expecially.

syedsajjadkazmii (commented):
Some thoughts/questions. Thankyou

syedsajjadkazmii (approved):

PR #1164, opened 2024-02-22
feat: simplify registration experiment

syedsajjadkazmii:
### Description

Adds a simplified registration experiment to authn registration page by moving the username field to the second step in the register page for users falling into variation.

#### JIRA

[VAN-1827](https://2u-internal.atlassian.net/browse/VAN-1827)

Locally

**Default Variation**

https://github.com/openedx/frontend-app-authn/assets/52817156/82f5b5a7-b64e-485e-aecc-6569255535d5

**Simplified Registration variation**

https://github.com/openedx/frontend-app-authn/assets/52817156/e16c8697-0705-4a8f-b180-cb47bef0e4c6

#### Merge Checklist

#### Post-merge Checklist


src/register/RegistrationPage.jsx
```diff
@@ -176,12 +218,15 @@ const RegistrationPage = (props) => {
   useEffect(() => {
     if (registrationResult.success) {
       // This event is used by GTM
-      sendTrackEvent('edx.bi.user.account.registered.client', {});
+      sendTrackEvent('edx.bi.user.account.registered.client', {
+        variation: simplifyRegistrationExpVariation,
+        picked_suggested_username: usernameSuggestionsBackup.includes(registrationResult?.authenticatedUser?.username),
```
ahtesham-quraish, line 223: I am not sure this ```registrationResult?.authenticatedUser?.username``` would return the username i guess we need to change it, correct me if I am wrong please

syedsajjadkazmii: I think the the authenticatedUser will have the username in the response. https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/user_authn/views/register.py#L593

src/register/data/optimizelyExperiment/helper.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export const prepareSimplifiedRegistrationFirstStepPayload = (
+  initPayload,
+  configurableFormFields,
+) => {
+  let payload = { ...initPayload };
+
+  // Removing username field that is in second step of registration form
+  delete payload.username;
```
mubbsharanwar, line 66: we are deleting the username from the payload, how are we handling this use case, calling API without a username?

src/register/data/utils.js
```diff
@@ -35,6 +35,7 @@ export const isFormValid = (
 ) => {
   const fieldErrors = { ...errors };
   let isValid = true;
+
```
mubbsharanwar, line 38: nit: remove it.

src/register/RegistrationFields/UsernameField/UsernameField.jsx
```diff
@@ -91,7 +91,7 @@ const UsernameField = (props) => {
     event.preventDefault();
     handleErrorChange('username', ''); // clear error
     handleChange({ target: { name: 'username', value: suggestion } }); // to set suggestion as value
-    dispatch(clearUsernameSuggestions());
+    dispatch(clearUsernameSuggestions(true));
```
ahtesham-quraish, line 94: why do we need to set true here?

mubbsharanwar (approved):
Looks good, but have some comments.

src/register/RegistrationFields/UsernameField/UsernameField.jsx
syedsajjadkazmii: Username suggestions are cleared when either the user clicks a suggestion of clicks clear button. We want to backup suggestions only when user clicks a suggestion therefore i introduced a temporary argument `isSuggestionClicked` in `clearUsernameSuggestion` and pass true only when the user clicked the suggestion. So that we can backup it.

src/register/data/optimizelyExperiment/helper.js
syedsajjadkazmii: This is just for validating complete first step of simplified register page. On first step, we don't have username so we are not validating it in the first step. We are validating username in the second step.

ahtesham-quraish (approved):

PR #1165, opened 2024-02-26
fix: Update custom SSO buttons to use brand colors

ahtesham-quraish:
### Description

Update custom SSO buttons to use brand colors 

#### JIRA

[VAN-1838](https://2u-internal.atlassian.net/browse/VAN-1838)

It has been tested locally

#### Screenshots

|  Before  |After|
|-----|-----|
|image| image|

### DIFF
--- a/src/sass/_style.scss
+++ b/src/sass/_style.scss
@@ -11,7 +11,6 @@
 // ----------------------------
 // #COLORS
 // ----------------------------
-$font-blue: #126f9a;
 $white: #FFFFFF;
 
 // social platforms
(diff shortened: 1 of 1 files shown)


src/sass/_style.scss
```diff
@@ -11,7 +11,7 @@
 // ----------------------------
 // #COLORS
 // ----------------------------
-$font-blue: #126f9a;
+$font-blue: #0e3639;
```
zainab-amir, line 14: Use primary color variable from paragon. Don't hardcode it.

zainab-amir: Also, this change doesn't make sense. Why does `font-blue` has some color that is not blue? 
You should update the SSO buttons to use `$primary` variable and not change it here. This variable may be used in some other CSS class and changing it here will impact those places too (If its not being used, remove the `font-blue` variable).

ahtesham-quraish: @zainab-amir this ```$font-blue``` is only used around sso button like color text, border color and background color that is why I have changed at one place instead of multiple places if you still think we need to changed it then we can change it

zainab-amir: Yes, logically it doesn't make sense to have green or yellow or grey color in "font-blue"

zainab-amir (approved):

PR #1168, opened 2024-02-26
fix: browser header showing null

attiyaIshaque:
### Description

When accessing Authn MFE for the first time it shows “Auth | null” in the browser header. In this PR we fix If process.env.SITE_NAME is not available it will show  “Authn” otherwise show “Authn | <site name>”

Backport PR: https://github.com/openedx/frontend-app-authn/pull/1170

#### JIRA

[VAN-1839](https://2u-internal.atlassian.net/browse/VAN-1839)

#### How Has This Been Tested?

Please describe in detail how you tested your changes.

#### Screenshots/sandbox (optional):

|Before|After|
|-------|-----|
|Screenshot 2024-02-26 at 3 53 42 PM|Screenshot 2024-02-26 at 4 02 39 PM|

Issue #1158, opened 2024-02-09
When accessing the Authn MFE first time, It shows "Authn | null" in the header


hinakhadim:
Whenever I clicked on 'Sign In' button, it takes me to the Authn MFE. The issue is that it shows 'Authn | null' in the header. The reason is that in `public/index.html`, the header is `<title>Authn | <%= process.env.SITE_NAME %></title>`. For first time, the `process.env.SITE_NAME` is null and after that `mergeConfig` runs which updates `SITE_NAME`. Due to that thing, it shows `Authn | null` which is not a good user experience. I will suggest to remove `| <%= process.env.SITE_NAME %>` and its better to just show `Authn` only in title.  Here is the screenshot:

<img width="1471" alt="Screenshot 2024-02-09 at 10 12 25 AM" src="https://github.com/openedx/frontend-app-authn/assets/112946189/6ace0fb0-e3fc-4417-ace8-fc0db4796fe0">

I am running openedx using `tutor local`

regisb:
Hi Hina! :wave: Removing the entire `| <%= process.env.SITE_NAME %>` is going to be a breaking change for every Open edX platform out there, and a very debatable one.

Can I suggest instead to only remove the site name when it's null? That way, the title would be "Authn" while loading the page, and then become "Authn | site_name" once `mergeConfig` is run. Is that even possible?

A separate question would be to rename the "Authn" title to "Authentication". I think there's consensus to say that "Authn" makes for a terrible site name.

attiyaIshaque:
@hinakhadim The issue has been resolved in Pull Request https://github.com/openedx/frontend-app-authn/pull/1168, and I am  closing the PR with the resolution of "Fixed."

attiyaIshaque:
@regisb I have implemented this suggested change.
> Can I suggest instead to only remove the site name when it's null? That way, the title would be "Authn" while loading the page, and then become "Authn | site_name" once `mergeConfig` is run. Is that even possible?
>

zainab-amir:
>  I think there's consensus to say that "Authn" makes for a terrible site name.

I agree. @attiyaIshaque lets create a follow up PR to do this too.

hinakhadim:
Hi all, I have tested the fixed PR #1168  on my machine. It haven't resolved the issue for me. It treats `null` as string and pass the condition which should be failed. I have created another PR #1185. I kindly request your assistance in reviewing this latest submission.

PR #1185, opened 2024-03-04
fix: browser header showing null and replace authn with Authentication

hinakhadim:
### Description

When accessing Authn MFE for the first time it shows "Auth | null" in the browser header. This PR will fix If process.env.SITE_NAME is not available it will show “Authentication” otherwise show “Authentication | {site_name}”. This bug  (#1158) has been fixed in #1168 but still not resolve. As `process.env.SITE_NAME` is `null` but in `index.html`, it assumes `null` as a string value and passed the condition which is expected to be failed. As a result, still showing `Authn | null`. Moreover, in this PR, I have replaced `Authn` with the `Authentication` as suggested [here](https://github.com/openedx/frontend-app-authn/issues/1158#issuecomment-1965850295). 

|Before|After|
|-------|-----|
|  auth2 | authn1|

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html lang="en-us">
   <head>
-      <title><%= process.env.SITE_NAME ? 'Authn | ' + process.env.SITE_NAME : 'Authn' %></title>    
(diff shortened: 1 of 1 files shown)


arbrandes:
I would've approved the fix, too, but at some point we should just start using `react-helmet` everywhere, like other MFEs [already do](https://github.com/openedx/frontend-app-learning/blob/master/src/index.jsx#L14).

PR #1194, opened 2024-03-05
feat: Remove simplify registration experiment code

attiyaIshaque:
### Description

We are presently conducting an A/B test (Hiding the username field from the registration form) on Authn. Following the conclusion of the experiment, our objective is to revert to the original state of the registration form. As part of this pull request, we are refactoring the code to eliminate the A/B experiment.

### NOTE

This change needs to be merged in the master on **March 12, 2024.** Please don’t merge before that as the experiment needs more time to run.

#### JIRA

[VAN-1859](https://2u-internal.atlassian.net/browse/VAN-1859)


src/register/RegistrationPage.test.jsx
```diff
@@ -104,6 +101,7 @@ describe('RegistrationPage', () => {
       registrationFormData,
       usernameSuggestions: [],
       usernameSuggestionsBackup: [],
```
syedsajjadkazmii, line 103: we need to remove this as well.

src/register/data/optimizelyExperiment/helper.js
```diff
@@ -1,131 +0,0 @@
-/**
```
syedsajjadkazmii, line 1: this file and its parent directory `optimizelyExperiment` can be removed.

syedsajjadkazmii (approved):

PR #1201, opened 2024-03-14
feat: remove username from the registration from

attiyaIshaque:
### Description

We've hidden the username field in the registration form when `ENABLE_AUTO_GENERATED_USERNAME`  feature flag is turned on. Additionally, we've excluded the username property from the registration request payload. 
A user is created successfully if a username is not provided.
#### JIRA

[VAN-1872](https://2u-internal.atlassian.net/browse/VAN-1872)

Tested locally 

|Before|After|
|-------|-----|
|Screenshot 2024-03-14 at 8 42 28 PM |Screenshot 2024-03-14 at 8 41 43 PM|

#### Merge Checklist

#### Post-merge Checklist


src/register/RegistrationPage.jsx
```diff
@@ -333,16 +337,18 @@ const RegistrationPage = (props) => {
...
-                name="username"
-                spellCheck="false"
-                value={formFields.username}
-                handleChange={handleOnChange}
-                handleErrorChange={handleErrorChange}
-                errorMessage={errors.username}
-                helpText={[formatMessage(messages['help.text.username.1']), formatMessage(messages['help.text.username.2'])]}
-                floatingLabel={formatMessage(messages['registration.username.label'])}
-              />
+              {!flags.autoGeneratedUsernameEnabled && (
```
syedsajjadkazmii, line 340: Can we do this?
```suggestion
              {!!flags.autoGeneratedUsernameEnabled && (
```

src/register/RegistrationPage.test.jsx
```diff
@@ -299,6 +306,31 @@ describe('RegistrationPage', () => {
       });
     });
 
+    it('should submit form without UsernameField when autoGeneratedUsernameEnabled is true', () => {
```
syedsajjadkazmii, line 309: can we add one more test that username field is not displayed when `ENABLE_AUTO_GENERATED_USERNAME` is true?

src/register/RegistrationPage.jsx
attiyaIshaque: We cannot do this because when `autoGeneratedUsernameEnabled` is true, we need to hide this. Using `!True `means false, which indicates that this code block should not execute.

src/register/RegistrationPage.test.jsx
attiyaIshaque: Sure Added.

syedsajjadkazmii (approved):

PR #1203, opened 2024-03-14
fix: useSelector usage in logistration and login component

syedsajjadkazmii:
Description
Optimization PR: This PR enhances the usage of useSelector in the Login and Logistration components, adhering to best practices. By doing so, it reduces unnecessary renders in these components, leading to improved performance.

Docs: Selectors should be as granular as possible
Reference: https://react-redux.js.org/api/hooks

JIRA
[VAN-1866](https://2u-internal.atlassian.net/browse/VAN-1866)

How Has This Been Tested?
Locally


src/login/LoginPage.jsx
```diff
@@ -46,28 +46,21 @@ import ResetPasswordSuccess from '../reset-password/ResetPasswordSuccess';
 const LoginPage = (props) => {
   const dispatch = useDispatch();
   const { formatMessage } = useIntl();
+  const backedUpFormData = useSelector(state => state.login.loginFormData);
+  const loginErrorCode = useSelector(state => state.login.loginErrorCode);
+  const loginErrorContext = useSelector(state => state.login.loginErrorContext);
+  const loginResult = useSelector(state => state.login.loginResult);
+  const shouldBackupState = useSelector(state => state.login.shouldBackupState);
+  const showResetPasswordSuccessBanner = useSelector(state => state.login.showResetPasswordSuccessBanner);
+  const submitState = useSelector(state => state.login.submitState);
```
mubbsharanwar, line 55: Have you noticed a decrease in rerendering the login component by granular selector state?

syedsajjadkazmii: There was an extra render of login component which was actually being caused by the re-rendering of its parent component (Logistration). 

Logistration was subscribed to register reducer thus a change in register redux state was causing the Logistration hence the Login component to re-render. 

By making the useSelectors granular in both components removed that extra rendered being caused by register store change.

PR #1207, opened 2024-03-26
use useSelector with granular approach in registration component to minimize rendering

mubbsharanwar:
### Description

In this PR we are optimizing the `registration` component by using reducer state more granular which helps to minimize the component rendering on dependent state change along with following best practices.

Docs: Selectors should be as granular as possible
Reference: https://react-redux.js.org/api/hooks
#### JIRA

[VAN-1890](https://2u-internal.atlassian.net/browse/VAN-1890)

Locally

### DIFF
--- a/src/register/RegistrationPage.jsx
+++ b/src/register/RegistrationPage.jsx
@@ -25,8 +25,9 @@ import {
   FORM_SUBMISSION_ERROR,
   TPA_AUTHENTICATION_FAILURE,
 } from './data/constants';
+import getBackendValidations from './data/selectors';
 import {
-  getBackendValidations, isFormValid, prepareRegistrationPayload,
+  isFormValid, prepareRegistrationPayload,
 } from './data/utils';
 import messages from './messages';
 import { EmailField, NameField, UsernameField } from './RegistrationFields';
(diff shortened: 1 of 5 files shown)


src/register/RegistrationPage.jsx
```diff
@@ -65,34 +65,25 @@ const RegistrationPage = (props) => {
...
-      finishAuthUrl,
-      currentProvider,
-      providers,
-      secondaryProviders,
-      pipelineUserDetails,
-    },
-  } = useSelector(state => state.commonComponents);
+  const backedUpFormData = useSelector(state => state.register.registrationFormData);
+  const registrationError = useSelector(state => state.register.registrationError);
+  const registrationErrorCode = useSelector(state => state.register.registrationError?.errorCode);
```
syedsajjadkazmii, line 70: It would get default value `undefined` if errorCode is not available. In the original implementation, the value will be `{}` if errorCode is not available. Can you check if the following work?
```suggestion
const registrationErrorCode = useSelector(state => state.register.registrationError?.errorCode ?? {});
```

src/register/RegistrationPage.jsx
```diff
@@ -65,34 +65,25 @@ const RegistrationPage = (props) => {
...
+  const registrationResult = useSelector(state => state.register.registrationResult);
+  const shouldBackupState = useSelector(state => state.register.shouldBackupState);
+  const userPipelineDataLoaded = useSelector(state => state.register.userPipelineDataLoaded);
+  const submitState = useSelector(state => state.register.submitState);
+  const validations = useSelector(state => state.register.submitState);
```
syedsajjadkazmii, line 75:
```suggestion
  const validations = useSelector(state => state.register.validations);
```

syedsajjadkazmii (changes requested):
Please confirm if its tested thoroughly

src/register/RegistrationPage.jsx
mubbsharanwar: In the original implementation, the default value is `{}` for `registrationError` not for `registrationErrorCode` and in our implementation the default value is `{}` too for `registrationError` but `undefined` for `registrationErrorCode` and it would be work fine because where the `registrationErrorCode` is using there is if condition in `useEffect` and it will handle the `undefined` case.

mubbsharanwar:
> Please confirm if its tested thoroughly

Yes, I tested it on success, failure, backend validations, username suggestions, etc cases but not tested the SSO case because I am unable to configure it locally.

syedsajjadkazmii (approved):

PR #1210, opened 2024-03-29
refactor: replace camelcase with snake case

mubbsharanwar:
### Description

Convert API response data from snake case to camel case for ES convention for code quality improvement.
**Note:**
In the future, we will handle API snake case response to the camel case in Django sterilizers and then directly use them in the authn MFE
#### JIRA

[VAN-1892](https://2u-internal.atlassian.net/browse/VAN-1892)

Locally


ahtesham-quraish (commented):
Did you test this locally extensively ?

src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
```diff
@@ -69,7 +69,7 @@ describe('ProgressiveProfilingTests', () => {
     },
   };
   const extendedProfile = ['company'];
-  const optionalFields = { fields, extended_profile: extendedProfile };
```
syedsajjadkazmii, line 72: You would need to update it in other tests as well.

src/progressive-profiling/ProgressiveProfiling.jsx
```diff
@@ -154,9 +154,9 @@ const ProgressiveProfiling = (props) => {
       'edx.bi.welcome.page.submit.clicked',
       {
         isGenderSelected: !!values.gender,
-        isYearOfBirthSelected: !!values.year_of_birth,
-        isLevelOfEducationSelected: !!values.level_of_education,
-        isWorkExperienceSelected: !!values.work_experience,
+        isYearOfBirthSelected: !!values.yearOfBirth,
+        isLevelOfEducationSelected: !!values.levelOfEducation,
+        isWorkExperienceSelected: !!values.workExperience,
```
syedsajjadkazmii, line 159: please check for these occurances in other places too.

For example the <RedirectLogistration> component rendering in this file.

syedsajjadkazmii (commented):
Please look for these occurances as well (in the tests files as well). 
Note: Please check if all these occurrences even need conversion atm.

- `field_name`
- `field_value`
- `level_of_education`
- `error_message` 
- `honor_code`
- `tos_and_honor_code`
- `social_auth_provider`
- `confirm_email`
- `marketing_emails_opt_in`
- `platform_name`
- `terms_of_service`


Thankyou

mubbsharanwar:
> * field_name

We updated all the possible occurrences which we can update without any dependency, and for `confirm_email`, `error_message` etc first we need to update in the API response and then in the components.

syedsajjadkazmii:
> > * field_name
> 
> We updated all the possible occurrences which we can update without any dependency, and for `confirm_email`, `error_message` etc first we need to update in the API response and then in the components.

Can we use camelCase function on API response in authn?

mubbsharanwar:
> Did you test this locally extensively ?

Yes!

mubbsharanwar:
> > > * field_name
> > 
> > 
> > We updated all the possible occurrences which we can update without any dependency, and for `confirm_email`, `error_message` etc first we need to update in the API response and then in the components.
> 
> Isn't the camelCase function (that we are using in service.js) making `confirm_email` -> `confirmEmail`?

No, the camelCase function just converts the first level dict object keys, and  `confirm_email` and `error_message` are inside fields

src/progressive-profiling/tests/ProgressiveProfiling.test.jsx
mubbsharanwar: Done.

syedsajjadkazmii:
> > > > * field_name
> > > 
> > > 
> > > We updated all the possible occurrences which we can update without any dependency, and for `confirm_email`, `error_message` etc first we need to update in the API response and then in the components.
> > 
> > 
> > Isn't the camelCase function (that we are using in service.js) making `confirm_email` -> `confirmEmail`?
> 
> No, the camelCase function just converts the first level dict object keys, and `confirm_email` and `error_message` are inside fields

Alright. So I would say lets discuss if we want to do this or not.

PR #1239, opened 2024-04-22
fix: fix register button width

syedsajjadkazmii:
### Description

Putting minimum width for btn to avoid de creasing width on loader.


mubbsharanwar (approved):
LGTM.

PR #1244, opened 2024-04-29
fix: add new entry for another US label

ahtesham-quraish:
Description: 

### Description

Add new entry for for another US label which is United States 

#### JIRA

[VAN-1900](https://2u-internal.atlassian.net/browse/VAN-1900)

It has been tested locally.

#### Screenshots:

|After|
|-----|
|image |

### DIFF
--- a/src/register/RegistrationFields/CountryField/CountryField.jsx
+++ b/src/register/RegistrationFields/CountryField/CountryField.jsx
@@ -97,7 +97,7 @@ const CountryField = (props) => {
   };
 
   const getCountryList = () => countryList.map((country) => (
-    <FormAutosuggestOption key={country[COUNTRY_CODE_KEY]} id={country[COUNTRY_CODE_KEY]}>
+    <FormAutosuggestOption key={country[COUNTRY_DISPLAY_KEY]} id={country[COUNTRY_CODE_KEY]}>
       {country[COUNTRY_DISPLAY_KEY]}
     </FormAutosuggestOption>
   ));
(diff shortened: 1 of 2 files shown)


src/register/RegistrationFields/CountryField/CountryField.jsx
```diff
@@ -97,7 +97,7 @@ const CountryField = (props) => {
   };
 
   const getCountryList = () => countryList.map((country) => (
-    <FormAutosuggestOption key={country[COUNTRY_CODE_KEY]} id={country[COUNTRY_CODE_KEY]}>
+    <FormAutosuggestOption key={country[COUNTRY_DISPLAY_KEY]} id={country[COUNTRY_CODE_KEY]}>
```
mubbsharanwar, line 100: Would it be fine if there are two same IDs? Would it work?

```
{
  code: 'US', name: 'United States of America',
  code: 'US', name: 'United States',
}
```

src/register/components/ConfigurableRegistrationForm.jsx
```diff
@@ -33,7 +33,10 @@ const ConfigurableRegistrationForm = (props) => {
     autoSubmitRegistrationForm,
   } = props;
 
-  const countryList = useMemo(() => getCountryList(getLocale()), []);
+  // The reason for adding the entry 'United States' is that sometime browser aut-fill the form with the 'Unites States'
+  // instead 'United States of America' which does not exists in country dropdown list and gets the user confused
+  // and unable to create account. So we added the United States entry in dropdown list.
```
mubbsharanwar, line 38:
```suggestion
 /** The reason for adding the entry 'United States' is that Chrome browser aut-fill the form with the 'Unites 
  States' instead of 'United States of America' which does not exist in country dropdown list and gets the user 
  confused and unable to create an account. So we added the United States entry in the dropdown list.
 */
```

PR #1245, opened 2024-05-06
revert: multistep registration experiment

mubbsharanwar:
### Description

We launched optimizely experiment for the multistep registration experience and after completing the experiment we are reverting the multistep registration experiment changes.

#### JIRA

[VAN-1930](https://2u-internal.atlassian.net/browse/VAN-1930)

We test it locally and using unit test coverage.

#### Merge Checklist

#### Post-merge Checklist


src/register/data/utils.js
```diff
@@ -43,39 +43,30 @@ export const isFormValid = (
...
-          isValid = false;
-        }
-        emailSuggestion = suggestion;
+      const {
+        fieldError, confirmEmailError, suggestion,
+      } = validateEmail(payload.email, configurableFormFields?.confirm_email, formatMessage);
+      if (fieldError) {
+        fieldErrors.email = fieldError;
+        isValid = false;
+      }
```
syedsajjadkazmii, line 56: i think the changes here resolves an existing bug. We should make a ticket for this.

syedsajjadkazmii (approved):

src/register/data/utils.js
mubbsharanwar: Which bug?

syedsajjadkazmii: he function does not check if the field already has an error and overrides the error with frontend validations. For example, if email field has an error of already existing email and we pass email field through isFormValid Function on submit, it will override the error with frontend validations (will only check if email is valid) and make submit call. Instead in actual, the call should not be made when we had an existing  email bug.

mubbsharanwar: Address!

PR #1257, opened 2024-06-13
build: Update codecov and use token

arbrandes:
Update codecov to the latest version and start using the org-wide token for uploads.

See https://github.com/openedx/wg-frontend/issues/179

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,4 +41,7 @@ jobs:
       run: npm run build
 
     - name: Run Code Coverage
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true


.github/workflows/ci.yml
```diff
@@ -41,4 +41,7 @@ jobs:
       run: npm run build
 
     - name: Run Code Coverage
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true
```
brian-smith-tcril, line 47: `fail_ci_if_error` is set to `false` in `frontend-app-account` (and unchanged by https://github.com/openedx/frontend-app-account/pull/1063/)

Just curious as to the reasoning behind explicitly setting it to `true` here.

arbrandes: The main motivation here is to follow the pattern given by https://openedx.atlassian.net/wiki/spaces/COMM/pages/3438280709/Adding+Codecov.  As to why the pattern is that way, I didn't write that page but I agree it makes sense to fail CI if coverage is not up to par, by default, so that at least the question is raised for every PR.  Exceptions can always be made when necessary.

I did not change the ones that were set to `false` explicitly, assuming there was a reason they were set like that.  We can always tackle that question separately.  (I would definitely like to understand when _not_ to fail CI.)

brian-smith-tcril (approved):

PR #1287, opened 2024-07-17
feat: onboarding component recommendations on authn

mubbsharanwar:
### Description

Show recommendations to the users who come from the onboarding component experience

#### JIRA

[VAN-1947](https://2u-internal.atlassian.net/browse/VAN-1947)

Through unit tests and manual testing


ahtesham-quraish (approved):
LGTM!

src/recommendations/RecommendationsPage.jsx
```diff
@@ -21,18 +22,35 @@ import RecommendationsLargeLayout from './RecommendationsPageLayouts/LargeLayout
...
+import { getAllPossibleQueryParams } from '../data/utils';
 
 const RecommendationsPage = () => {
   const { formatMessage } = useIntl();
+  const DASHBOARD_URL = getConfig().LMS_BASE_URL.concat(DEFAULT_REDIRECT_URL);
   const isExtraSmall = useMediaQuery({ maxWidth: breakpoints.extraSmall.maxWidth - 1 });
   const location = useLocation();
+  const queryParams = getAllPossibleQueryParams();
+  // flag to show recommendations for onboarding component experience
+  const showRecommendations = !!queryParams?.levelOfEducation || !!queryParams?.finishAuthUrl || !!queryParams?.country;
```
syedsajjadkazmii, line 34: should this be `&&` instead of `||`?

src/recommendations/RecommendationsPage.jsx
```diff
@@ -46,8 +64,8 @@ const RecommendationsPage = () => {
 
   const handleSkipRecommendationPage = () => {
     window.history.replaceState(location.state, null, '');
-    if (registrationResponse) {
-      window.location.href = registrationResponse.redirectUrl;
+    if (redirectUrl) {
+      window.location.href = redirectUrl;
     } else {
       window.location.href = DASHBOARD_URL;
     }
```
syedsajjadkazmii, line 71: this would disturb the authn mfe registration flow and will not redirect users to redirect URL after the user registered from authn mfe.

src/recommendations/RecommendationsPage.jsx
```diff
@@ -59,7 +77,7 @@ const RecommendationsPage = () => {
     setTimeout(() => { handleSkipRecommendationPage(); }, LINK_TIMEOUT);
   };
 
-  if (!registrationResponse) {
+  if (!redirectUrl && !showRecommendations) {
```
syedsajjadkazmii, line 80: same 
> this would disturb the authn mfe registration flow

syedsajjadkazmii (changes requested):

src/recommendations/RecommendationsPage.jsx
mubbsharanwar: levelOfEducation can be null if user not select the option from progressive profiling, also finishAuthUrl can be null.

src/recommendations/RecommendationsPage.jsx
mubbsharanwar: How?
This wouldn't disturb the authn mfe registration flow because if there is no `redirectUrl` and the user is not coming onboarding experience then It should be redirected to the dashboard, and if redirectUrl exist user will able to see the recommendations

src/recommendations/RecommendationsPage.jsx
mubbsharanwar: This will not disturb the authn MFE flow because we introduce a state and set `registrationResponse.redirectUrl` to `redirectUrl` as a initial state value and using it here in if condation.

src/recommendations/RecommendationsPage.jsx
```diff
@@ -21,18 +22,36 @@ import RecommendationsLargeLayout from './RecommendationsPageLayouts/LargeLayout
...
 
 const RecommendationsPage = () => {
   const { formatMessage } = useIntl();
+  const DASHBOARD_URL = getConfig().LMS_BASE_URL.concat(DEFAULT_REDIRECT_URL);
   const isExtraSmall = useMediaQuery({ maxWidth: breakpoints.extraSmall.maxWidth - 1 });
   const location = useLocation();
+  const queryParams = getAllPossibleQueryParams();
+  // flag to show recommendations for onboarding component experience
+  const showRecommendations = !!queryParams?.levelOfEducation
+    || (!!queryParams?.finalRedirectUrl && !!queryParams?.country);
```
syedsajjadkazmii, line 35: this would show recommendations even if there is only level of education in query params. I think we should have some other checks as well. maybe we can introduce a new query param? `referrer=onboarding-component`?

src/recommendations/RecommendationsPage.jsx
```diff
@@ -21,18 +22,36 @@ import RecommendationsLargeLayout from './RecommendationsPageLayouts/LargeLayout
...
+  const [educationLevel, setEducationLevel] = useState(EDUCATION_LEVEL_MAPPING[location.state?.educationLevel]);
+  const [userId, setUserId] = useState(location.state?.userId || -1);
+  const [userCountry, setUserCountry] = useState(backendCountryCode);
+
+  useEffect(() => {
+    if (showRecommendations) {
+      const authenticatedUser = getAuthenticatedUser();
+      if (authenticatedUser) {
+        setRedirectUrl(queryParams.finalRedirectUrl);
+        setEducationLevel(EDUCATION_LEVEL_MAPPING[queryParams.levelOfEducation]);
```
syedsajjadkazmii, line 48: as level of education is optional, this would crash on `queryParams.levelOfEducation`.

src/recommendations/RecommendationsPage.jsx
syedsajjadkazmii: finishAuthUrl can't be null, it will be `dashboard` in fallback/default case.

src/recommendations/RecommendationsPage.jsx
syedsajjadkazmii: you are right.

src/recommendations/RecommendationsPage.jsx
syedsajjadkazmii: you are right.

src/recommendations/RecommendationsPage.jsx
mubbsharanwar: The condation is with `OR` operator with `levelOfEducation` and (finalRedirectUrl && country), so if there is only level of education in query params the recommendations will be shown.

syedsajjadkazmii (approved):

src/recommendations/RecommendationsPage.jsx
```diff
@@ -46,8 +64,8 @@ const RecommendationsPage = () => {
 
   const handleSkipRecommendationPage = () => {
     window.history.replaceState(location.state, null, '');
-    if (registrationResponse) {
-      window.location.href = registrationResponse.redirectUrl;
+    if (redirectUrl) {
+      window.location.href = redirectUrl;
```
syedsajjadkazmii, line 68: something to discuss before merging: 

I think it's a security risk to redirect to `redirectUrl` coming in query params without verifying if its safe or not.

syedsajjadkazmii (changes requested):
blocking merge until we discuss the above ^ comment.

PR #1300, opened 2024-08-20
fix: password reset redirection

mubbsharanwar:
### Description

fix authenticated user redirects to 404 if the token is invalid for password reset

#### JIRA

[VAN-2052](https://2u-internal.atlassian.net/browse/ VAN-2052)

tested through local testing


mubbsharanwar:
> There can be query params. Would this still work? Also dont we want to redirect to dashboard?

We are adding a condition for `RESET_PAGE` all other cases are with query params.
Also, we are redirecting to the forget password page because unauthenticated users are redirected to the forget password page if the token is invalid.

syedsajjadkazmii (approved):

PR #1320, opened 2024-09-20
feat: add MainAppSlot for chatbot plugin

awais-ansari:
[INF-1573](https://2u-internal.atlassian.net/browse/INF-1573)

### Description

- This PR adds MainAppSlot, allowing a plugin at the root level. For example the chatbot plugin.
- This PR removes the automatic use of the Zendesk component from the App level. The Zendesk logic still lives in the src/common-components directory.


ayesha-waris (approved):

asadazam93 (approved):

AhtishamShahid (approved):

jsnwesson:
This looks great so far! One thing that I would request is that in your `/plugin-slots` directory if you can follow the style of documentation that is in [Learner Dashboard ](https://github.com/openedx/frontend-app-learner-dashboard/tree/master/src/plugin-slots)and[ Learning](https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots)'s `/plugin-slots` directories. 
Both also include [a section in the main README](https://github.com/openedx/frontend-app-learning?tab=readme-ov-file#plugins) that plugin slots exist and points to where they can be found. That would be useful to include here as well.

jsnwesson:
Oh that's riiight, this is only a PR that merges into the forked `2u-main` branch. Disregard my request unless you think it'll be useful for your own team. If there's a future case where this PR will eventually be pushed up to the `master` branch, then I think that documentation will have a stronger ask.

jsnwesson (approved):

PR #1381, opened 2024-11-27
fix: fixed duplicate event firing on post SSO login registration

ayesha-waris:
[INF-1721](https://2u-internal.atlassian.net/browse/INF-1721)

**Description**

fix cohesion duplicate events post SSO login/registration


awais-ansari (approved):
Please update test case according to new condition.

PR #1611, opened 2025-12-08
feat!: add design tokens support

arbrandes:
### Description

See parent issue for details: https://github.com/openedx/frontend-app-authn/issues/1619

BREAKING CHANGE: Pre-design-tokens theming is no longer supported.

Depends on https://github.com/openedx/frontend-base/pull/111 and https://github.com/openedx/frontend-base/pull/115.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,8 +41,11 @@
         "react-test-renderer": "^18.3.1",
         "ts-jest": "^29.4.0"
       },
+      "engines": {
+        "node": "^24.12"
+      },
       "peerDependencies": {
-        "@openedx/frontend-base": "^1.0.0-alpha.7",
+        "@openedx/frontend-base": "^1.0.0-alpha.8",
         "@openedx/paragon": "^23",
         "react": "^18",
         "react-dom": "^18",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "@openedx/frontend-app-authn",
   "version": "1.0.0-alpha.4",
   "description": "Frontend authentication",
+  "engines": {
+    "node": "^24.12"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/openedx/frontend-app-authn.git"


src/sass/_style.scss
```diff
@@ -42,7 +40,7 @@ $elevation-level-2-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.15);
 }
 
 .main-content {
-  @extend .pt-4;
+  padding-top: calc(var(--pgn-spacing-spacer-base) * 1.5) !important;
```
brian-smith-tcril: This appears to also be using this verbose form [on `master`](https://github.com/openedx/frontend-app-authn/blob/master/src/sass/_style.scss#L43), but this can be cleaned up to use [`--pgn-spacing-spacer-4`](https://github.com/openedx/paragon/blob/9e30e8cdb211b94eccc39b24c810a893f27d0684/styles/css/core/variables.css#L545)

```css
--pgn-spacing-spacer-4: calc(var(--pgn-spacing-spacer-base) * 1.5); /** Space value of level 4 */
```

src/data/utils/useMobileResponsive.js
```diff
@@ -10,7 +10,7 @@ import { breakpoints } from '@openedx/paragon';
 const useMobileResponsive = (breakpoint) => {
   const [isMobileWindow, setIsMobileWindow] = useState();
   const checkForMobile = () => {
-    setIsMobileWindow(window.matchMedia(`(max-width: ${breakpoint || breakpoints.small.maxWidth}px)`).matches);
+    setIsMobileWindow(window.matchMedia(`(max-width: ${breakpoint ?? breakpoints.small.maxWidth}px)`).matches);
```
brian-smith-tcril: Looks like there are quite a few spots in this PR replacing `||` with `??`. I have the same comment for all of those as https://github.com/openedx/frontend-app-authn/pull/1611/changes#r2682650606. I'm all for cleaning this stuff up, but it'd be good to have it as a separate PR so we aren't "sneaking it in."

brian-smith-tcril:
Re: https://github.com/openedx/frontend-app-authn/pull/1611#pullrequestreview-3651190487

Since we found this is an issue on `24.9` but not `24.12` or `24.13`, a reasonable improvement for now is to ensure we get `EBADENGINE` warnings for node versions below `24.12`

```diff
diff --git a/package.json b/package.json
index c770f4e..29a2ce4 100644
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "@openedx/frontend-app-authn",
   "version": "1.0.0-alpha.4",
   "description": "Frontend authentication",
+  "engines": {
+    "node": "^24.12"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/openedx/frontend-app-authn.git"
```

PR #1627, opened 2026-02-05
feat: added code spliting on main module

holaontiveros:
### Description

This adds the Main path of auth as a lazy loaded component so any app consuming this (as frontend-template-site) will have the related routes with code splitting

fixes https://github.com/openedx/frontend-base/issues/161

- Add this repo as part of a parent aggregator project (use frontend-template-site as a quick headstart)
- Run the the frontent-base app and see that while accessing the related routes extra chunks of code are being fetch
- Pull any other route of this frontend-app if any work it should be the same for any other

#### Merge Checklist

* [ X ] Is there adequate test coverage for your changes? (if the actual tests work, then it's fine given that it's a loading related thing it either works or not)

#### Post-merge Checklist

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -1,5 +1,3 @@
-import { Navigate } from 'react-router';
-
 import { EmbeddedRegistrationRoute, NotFoundPage, UnAuthOnlyRoute } from './common-components';
 import { LOGIN_PAGE } from './data/constants';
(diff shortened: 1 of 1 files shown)


arbrandes (changes requested):
It looks like this works perfectly, thank you!  Can you just make a couple of changes, please?

PR #1628, opened 2026-02-06
feat!: compile to JS before publishing

arbrandes:
Configure the package to compile TypeScript and copy SCSS files to `dist/` before publishing, rather than publishing raw source files. This allows us to use tsc-alias for `@src` imports.

Also use a more modern export map to decouple the internal file structure from the package's API, and add a build step to CI.

#### BREAKING CHANGE

Consuming projects may need to update their imports or SASS `@use` lines.

#### Depends on

https://github.com/openedx/frontend-base/pull/164

#### LLM Usage Notice

Built with assistance from Claude Opus 4.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,9 @@ jobs:
     - name: Test
       run: npm run test
 
+    - name: Build
+      run: npm run build
+
     - name: Run Code Coverage
       uses: codecov/codecov-action@v5
       with:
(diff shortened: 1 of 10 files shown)


brian-smith-tcril:
> @brian-smith-tcril, I may have accidentally included the temporary release stuff in here, but it's [completely separate](https://github.com/openedx/frontend-base/pull/172) now.

The commits tab is still showing https://github.com/openedx/frontend-app-authn/pull/1628/changes/d3c422a98c1332dbcae3d99bc32d99cb2cd1a202

PR #1633, opened 2026-02-13
chore: update route base path

holaontiveros:
### Description

This will make the base route to match the old paths, which will alow to switch between MFE / frontendapps easily


arbrandes (changes requested):
Can you also update site.config.test.tsx with the same change, please?  After that we can merge this.

arbrandes (approved):
Ok, thanks!

PR #1663, opened 2026-04-09
feat: hide shell header and footer dynamically

### LINKED ISSUE #219: Header slot deactivation via `provides:`
### Description

Header slot deactivation is currently [hardcoded](https://github.com/openedx/frontend-base/blob/main/shell/app.ts#L6-L12).  Modify the mechanism so that apps can provide a list of roles that should deactivate the header slot via the new `provides:` mechanism.

arbrandes:
### Description

Uses the new `provides` mechanism from frontend-base to declare which authn roles should trigger chromeless mode (no shell header or footer). All five authn roles (login, register, resetPassword, confirmPassword, welcome) are listed under `providesChromelessRolesId`, preserving the existing behavior where header and footer are hidden on all authn pages.

Depends on openedx/frontend-base#220.

### LLM usage notice

Built with assistance from Claude.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4395,9 +4395,9 @@
       }
     },
     "node_modules/@openedx/frontend-base": {
-      "version": "1.0.0-alpha.23",
-      "resolved": "https://registry.npmjs.org/@openedx/frontend-base/-/frontend-base-1.0.0-alpha.23.tgz",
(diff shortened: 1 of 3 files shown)


brian-smith-tcril (approved):
LGTM!

<details>

<summary>What I was in the middle of writing before this updated</summary>

Once this includes a `package-lock.json` update the
```
Module '"@openedx/frontend-base"' has no exported member 'providesChromelessRolesId'.
```
problem should go away.

</details>

PR #1668, opened 2026-04-22
chore: add build:ci smoke test to verify webpack build

arbrandes:
### Description

Apps are distributed build-less, so `npm run build` only compiles the library `dist/` via `tsc` and nothing in CI verifies that the app can actually be webpack-bundled as a deployable site. This PR adds an `npm run build:ci` script that exercises the real app graph through webpack, and wires it into the CI workflow after the existing build step.

Refs openedx/frontend-base#124.

### Verifying the check

Introduce a typo in an import in `src/routes.jsx` (for example, change `from './forgot-password'` to `from './forgot-password-typo'`). `npm run build` still passes, but `npm run build:ci` fails with a module resolution error. Revert the typo and the build goes green again.

### LLM usage notice

Built with assistance from Claude.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,9 @@ jobs:
     - name: Build
       run: npm run build
 
+    - name: Build (CI)
+      run: npm run build:ci
+
     - name: Run Code Coverage
       uses: codecov/codecov-action@v5
       with:
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,9 @@ build:
 	    cp "$$f" "$$d"; \
 	  done' sh {} +
 
+build-ci:
+	SITE_CONFIG_PATH=site.config.ci.tsx openedx build
+
 extract_translations: | requirements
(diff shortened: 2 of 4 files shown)


site.config.dev.tsx
```diff
@@ -13,7 +13,12 @@ const siteConfig: SiteConfig = {
   logoutUrl: 'http://local.openedx.io:8000/logout',
 
   environment: EnvironmentTypes.DEVELOPMENT,
-  apps: [authnApp],
+  apps: [
+    shellApp,
+    headerApp,
+    footerApp,
+    authnApp,
+  ],
```
brian-smith-tcril, line 21: I assume the goal here is to verify the chromeless stuff from https://github.com/openedx/frontend-base/pull/220 works?

Is that something we need in the dev config?

arbrandes: It's not something we need, but something we want: we should catch during development whether chromeless is failing, for whatever reason.

brian-smith-tcril (approved):

